PR #38, opened 2015-09-30
Remove wrong diagram in docs

### LINKED ISSUE #9: Wrong diagram in docs
The "architecture diagram" under https://github.com/osiam/osiam/blob/master/docs/OSIAM-Overview.md#gateways-between-these-systems is not up-to-date and must be changed. IIRC it was initially done by @leberKleber, so one should get the source from him and update it.

@leberKleber do you remember this and could you provide the source file to us?

dacrome:
Closes #9


fwilhe:
Thanks for fixing this right away.

PR #53, opened 2015-10-24
Migrate to BCrypt hashes

### LINKED ISSUE #34: Migrate to BCrypt hashes.
Spring Security deprecated the `PasswordEncoder` interface for quite some time now. I suggest that we migrate the used password encoder to an BCrypt implementation. We should do that in two steps:
1. Provide our own implementation of a `PasswordEncoder` that proxies calls to the `isPaswordValid()` method and updates the password field in the database with the newly encrypted password on successful login.
2. Announce that we are going to lock out all users with the old password field after two releases (or a yet undefined length of time)
3. Replace our proxy implementation with the actual `BcryptEncoder` implementation after a clearly communicated grace period and a later release. The Flyway database migration for that release should lock all users that still use the old hashing method.

So, what do you think?

dacrome:
When users login, their passwords will be updated to the new BCrypt hashes.
Until the Release 4.0 the old sha password hashes will be supported.
After OSIAM 4.0 was released the old password hashes are not usable anymore.

Resolves #34

### DIFF
--- a/src/main/java/org/osiam/auth/login/internal/InternalAuthenticationProvider.java
+++ b/src/main/java/org/osiam/auth/login/internal/InternalAuthenticationProvider.java
@@ -42,6 +42,7 @@ import org.springframework.security.authentication.event.AuthenticationSuccessEv
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.GrantedAuthority;
 import org.springframework.security.core.authority.SimpleGrantedAuthority;
+import org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder;
 import org.springframework.stereotype.Component;
 
 import java.util.ArrayList;
@@ -60,17 +61,20 @@ public class InternalAuthenticationProvider implements AuthenticationProvider,
     private final Map<String, Date> lastFailedLogin = Collections.synchronizedMap(new HashMap<>());
 
     private final SCIMUserProvisioning userProvisioning;
-    private final ShaPasswordEncoder passwordEncoder;
+    private final ShaPasswordEncoder shaPasswordEncoder;
+    private final BCryptPasswordEncoder bCryptPasswordEncoder;
     private final Integer maxLoginFailures;
     private final Integer lockTimeout;
 
     @Autowired
     public InternalAuthenticationProvider(SCIMUserProvisioning userProvisioning,
-                                          ShaPasswordEncoder passwordEncoder,
+                                          ShaPasswordEncoder shaPasswordEncoder,
+                                          BCryptPasswordEncoder bCryptPasswordEncoder,
                                           @Value("${osiam.tempLock.count:0}") Integer maxLoginFailures,
                                           @Value("${osiam.tempLock.timeout:0}") Integer lockTimeout) {
         this.userProvisioning = userProvisioning;
-        this.passwordEncoder = passwordEncoder;
+        this.shaPasswordEncoder = shaPasswordEncoder;
+        this.bCryptPasswordEncoder = bCryptPasswordEncoder;
         this.maxLoginFailures = maxLoginFailures;
         this.lockTimeout = lockTimeout;
     }
@@ -93,7 +97,6 @@ public class InternalAuthenticationProvider implements AuthenticationProvider,
 
         assertUserNotLocked(username);
 
-        // Determine username
         User user = userProvisioning.getByUsernameWithPassword(username);
 
         if (user == null) {
@@ -104,8 +107,13 @@ public class InternalAuthenticationProvider implements AuthenticationProvider,
             throw new DisabledException("The user with the username '" + username + "' is disabled!");
         }
 
-        if (!passwordEncoder.isPasswordValid(user.getPassword(), password, user.getId())) {
-            throw new BadCredentialsException("Bad credentials");
+        if (!bCryptPasswordEncoder.matches(password, user.getPassword())) {
+            if (shaPasswordEncoder.isPasswordValid(user.getPassword(), password, user.getId())) {
+                User replaceUser = new User.Builder(user).setPassword(password).build();
+                userProvisioning.replace(user.getId(), replaceUser);
+            } else {
+                throw new BadCredentialsException("Bad credentials");
+            }
         }
 
         User authUser = new User.Builder(username).setId(user.getId()).build();
--- a/src/main/java/org/osiam/resources/provisioning/update/UserUpdater.java
+++ b/src/main/java/org/osiam/resources/provisioning/update/UserUpdater.java
@@ -30,7 +30,7 @@ import org.osiam.resources.scim.User;
 import org.osiam.storage.dao.UserDao;
 import org.osiam.storage.entities.UserEntity;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.security.authentication.encoding.PasswordEncoder;
+import org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder;
 import org.springframework.stereotype.Service;
 
 import java.util.HashSet;
@@ -53,7 +53,7 @@ public class UserUpdater {
     private final RoleUpdater roleUpdater;
     private final X509CertificateUpdater x509CertificateUpdater;
     private final AddressUpdater addressUpdater;
-    private final PasswordEncoder passwordEncoder;
+    private final BCryptPasswordEncoder bCryptPasswordEncoder;
     private final ExtensionUpdater extensionUpdater;
 
     @Autowired
@@ -68,7 +68,7 @@ public class UserUpdater {
                        RoleUpdater roleUpdater,
                        X509CertificateUpdater x509CertificateUpdater,
                        AddressUpdater addressUpdater,
-                       PasswordEncoder passwordEncoder,
+                       BCryptPasswordEncoder bCryptPasswordEncoder,
                        ExtensionUpdater extensionUpdater) {
         this.userDao = userDao;
         this.resourceUpdater = resourceUpdater;
@@ -81,7 +81,7 @@ public class UserUpdater {
         this.roleUpdater = roleUpdater;
         this.x509CertificateUpdater = x509CertificateUpdater;
         this.addressUpdater = addressUpdater;
-        this.passwordEncoder = passwordEncoder;
+        this.bCryptPasswordEncoder = bCryptPasswordEncoder;
         this.extensionUpdater = extensionUpdater;
     }
 
@@ -234,7 +234,7 @@ public class UserUpdater {
         }
 
         if (!Strings.isNullOrEmpty(user.getPassword())) {
-            String hashedPassword = passwordEncoder.encodePassword(user.getPassword(), userEntity.getId());
+            String hashedPassword = bCryptPasswordEncoder.encode(user.getPassword());
             userEntity.setPassword(hashedPassword);
         }
     }
--- a/src/test/groovy/org/osiam/auth/login/internal/InternalAuthenticationProviderSpec.groovy
+++ b/src/test/groovy/org/osiam/auth/login/internal/InternalAuthenticationProviderSpec.groovy
@@ -23,15 +23,96 @@
  */
 package org.osiam.auth.login.internal
 
+import org.osiam.resources.provisioning.SCIMUserProvisioning
+import org.osiam.resources.scim.User
+import org.springframework.security.authentication.BadCredentialsException
+import org.springframework.security.authentication.encoding.ShaPasswordEncoder
+import org.springframework.security.core.Authentication
+import org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder
+import spock.lang.Shared
 import spock.lang.Specification
 
 class InternalAuthenticationProviderSpec extends Specification {
 
-    def 'the internal provider only supports InternalAuthentication class'() {
-        given:
-        InternalAuthenticationProvider provider = new InternalAuthenticationProvider(null, null, null, null)
+    // hashs of 'password'
+    def VALID_BCRYPT_HASHED_PASSWORD = '$2a$13$PfXyvWHogpPZgZjnk2a0eO7wSSNsMZ6DL4x5k8Bk0CK1cwXwMs0de'
+    def VALID_SHA_HASHED_PASSWORD = 'dc9b4c4a06e7dc774e469a16243ff448a02932fd323bb7ecb0ece308d6bb13c964dd7a0b7b99b15dc785004bdb869450f4070a9f83b027af43025149d1bc8f34'
 
+    // hash of 'wrongpassword'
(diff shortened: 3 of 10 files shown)


dacrome:
Sorry, forgot to comment that this PR is WIP.
We need to talk about: 
- How long we want to support both encoder, stop after two releases is what @wallner [suggested](https://github.com/osiam/osiam/issues/34#issue-108123652)
- The [strength of the gensalt](https://github.com/osiam/osiam/pull/53/files#diff-2f75936ef4808d8d3565bd0623f54a2fR100), I chose 12, based on recommendation I read somewhere (sorry, post the link later)
- How we deal with old hashes after the support for the old encoder ends, @wallner [suggested](https://github.com/osiam/osiam/issues/34#issue-108123652) to lock the users with a Flyway migration

Corresponding ITs https://github.com/osiam/connector4java-integration-tests/pull/305

tkrille:
> How long we want to support both encoder, stop after two releases is what @wallner suggested

"after two releases" a little bit ambiguous. Does it mean we remove support for old hashed passwords in OSIAM 3.3 or 3.4?

> The strength of the gensalt, I chose 12, based on recommendation I read somewhere (sorry, post the link later)

12 is the currently recommended default. Or is it 13 already? I think this should be made configurable. After all BCrypt is designed to get slower over time to compensate increasing computing power. Let's start with 12 as the default value and revise the default every year. Is it possible to change the value without invalidating all current passwords, i.e. is the number of rounds stored with the hash?

> How we deal with old hashes after the support for the old encoder ends, @wallner suggested to lock the users with a Flyway migration

They're locked anyway, because there's no way to generate the old hash anymore. I think users that missed the grace period should just reset their password or create a new account, ore something else, depending on the application. But why bother the users of OSIAM which such hassle?

At this time we store salted SHA-512 hashed 1000 times. Why not use these hashes to bootstrap BCrypt? We could just create a combined password encoder that will work forever by first hashing the password with SHA-512 a thousand times before passing it to BCrypt. And a thousand times isn't really much. My Laptop makes 1200 rounds each millisecond. So it would not even bother the server's CPU to calculate this hash for each login. The actual migration to a BCrypy hash can then be made automatically (flyway or other migration) or when the user logs in the next time (like planned).

> The search needs to be updated, the password should not be recognized

Do you mean this: #56?

tkrille:
Another question: Is this scheduled for 3.0?

dacrome:
Yes. Any comments on this?

tkrille:
> Yes. Any comments on this?

I don't understand the question.

dacrome:
> I don't understand the question.

Sorry, I meant do you have any questions about this PR changes?

tkrille:
Yes, I do: https://github.com/osiam/osiam/pull/53#issuecomment-160771026 :)

src/main/java/org/osiam/resources/provisioning/update/UserUpdater.java
```diff
@@ -33,6 +33,7 @@
 import org.osiam.storage.entities.UserEntity;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.security.authentication.encoding.PasswordEncoder;
```
wallner: Codacy Issue found: [Avoid unused imports such as 'org.springframework.security.authentication.encoding.PasswordEncoder'](https://www.codacy.com/app/OSIAM/osiam/file/1842979935/issues/source?bid=3080997&fileBranchId=3081607#l35)

wallner: Codacy Issue found: [Avoid unused imports such as 'org.springframework.security.authentication.encoding.PasswordEncoder'](https://www.codacy.com/app/OSIAM/osiam/file/2025501359/issues/source?bid=3080997&fileBranchId=3081607#l35)

dacrome:
> "after two releases" a little bit ambiguous. Does it mean we remove support for old hashed passwords in OSIAM 3.3 or 3.4?

Yes, that's ambiguous. What about to remove it with OSIAM 4.0?

> 12 is the currently recommended default. Or is it 13 already? I think this should be made configurable. After all BCrypt is designed to get slower over time to compensate increasing computing power. Let's start with 12 as the default value and revise the default every year. Is it possible to change the value without invalidating all current passwords, i.e. is the number of rounds stored with the hash?

[I guess it's possible](http://crypto.stackexchange.com/questions/3003/do-i-have-to-recompute-all-hashes-if-i-change-the-work-factor-in-bcrypt), but I need to spend more time on that. I don't know if the rounds stored, but I check this.

> They're locked anyway, because there's no way to generate the old hash anymore. I think users that missed the grace period should just reset their password or create a new account, ore something else, depending on the application. But why bother the users of OSIAM which such hassle?

I hope that we provide a password reset functionality when integrating the self-administration in the core service ;)

> At this time we store salted SHA-512 hashed 1000 times. Why not use these hashes to bootstrap BCrypt? We could just create a combined password encoder that will work forever by first hashing the password with SHA-512 a thousand times before passing it to BCrypt. And a thousand times isn't really much. My Laptop makes 1200 rounds each millisecond. So it would not even bother the server's CPU to calculate this hash for each login. The actual migration to a BCrypy hash can then be made automatically (flyway or other migration) or when the user logs in the next time (like planned).

@wallner mentioned that it is not a good idea to hash a hash. Again, also to this topic I ask google.

> > The search needs to be updated, the password should not be recognized
> 
> Do you mean this: #56?

Yes, I updated the todo list in [my comment](https://github.com/osiam/osiam/pull/53#issuecomment-150986916)

dacrome:
Updated and rebased

wallner:
Doing a little bit of research I found [this paper](https://eprint.iacr.org/2015/070) about combining hashing algorithms and comes to the conclusion that combining hash algorithms might not be the best idea. I realize, that this is something other than what @tkrille has in mind, however, I would still recommend avoiding mixing hash-alogrithms.

src/main/java/org/osiam/auth/login/internal/InternalAuthenticationProvider.java
```diff
@@ -104,8 +107,13 @@ public Authentication authenticate(Authentication authentication) {
             throw new DisabledException("The user with the username '" + username + "' is disabled!");
         }
 
-        if (!passwordEncoder.isPasswordValid(user.getPassword(), password, user.getId())) {
-            throw new BadCredentialsException("Bad credentials");
+        if (!bCryptPasswordEncoder.matches(password, user.getPassword())) {
+            if (passwordEncoder.isPasswordValid(user.getPassword(), password, user.getId())) {
+                User replaceUser = new User.Builder(user).setPassword(password).build();
+                userProvisioning.replace(user.getId(), replaceUser);
+            } else {
```
wallner, line 114: I think the `else` block needs to be one level out.

dacrome, line 114: Then the `BadCredentialsException` would be thrown when the password matches with the bcrypt password encoder.
The `BadCredentialsException` is thrown when both encoders do not return with true.

src/main/java/org/osiam/auth/login/internal/InternalAuthenticationProvider.java
```diff
@@ -61,16 +62,19 @@
 
     private final SCIMUserProvisioning userProvisioning;
     private final ShaPasswordEncoder passwordEncoder;
```
wallner: I would suggest that you rename the old `passwordEncoder` to `shaPasswordEncoder`, to make obvious what's happening

dacrome: Ok

dacrome:
Updated

dacrome:
Before this is mergeable I need to check what the recommend default for the strength of the gensalt

tkrille:
@wallner I did some research, too. Seems that somehow combining hash functions is not a problem at all, but the way they are combined is. Neither simple concatenating nor composition provide sufficient security. Thanks to the paper you've linked, XOR'ing has been finally proven insecure. There seems some new and exciting ways to securely combine hash functions, but I don't think it's worth the effort. After all, BCrypt is already secure enough and we don't strive to make it more secure by pre-hashing passwords, we only want to keep the migration efforts to a minimum. There are also a bunch of problems with losing entropy, depending on the password lengths, key sizes and the hash function's output format.

So long story short, we should stick with the initial idea: migrate the users to bcrypt and throw the old hashes away.

tkrille:
@dacrome I think it's 13 nowadays.

wallner:
So, @dacrome I'm fine with this PR and will happily merge it, but as far as I can tell it's missing tests for the crucial logic of it. Please add some.

dacrome:
@tkrille ok, I will update to the gensalt strength of 13... sadly, I have to update all database seeds in the ITs

@wallner damn, off course I have to!

dacrome:
@tkrille updated strength of gensalt to 13, also [all database seeds in the IT](https://github.com/osiam/connector4java-integration-tests/pull/305)

dacrome:
@wallner I add some unit tests later. I can think of mainly test the [change in the InternalAuthenticationProvider](https://github.com/osiam/osiam/pull/53/files#diff-420d799ac92afb23c8234ec4a8ecad73R110), right?

wallner:
Yes. Please.

dacrome:
@wallner [added some unit tests](https://github.com/osiam/osiam/pull/53/files#diff-2ce849f09a9760d4ef3024be491cd7d2R34)

src/test/groovy/org/osiam/auth/login/internal/InternalAuthenticationProviderSpec.groovy
```diff
@@ -23,15 +23,96 @@
...
+    def VALID_SHA_HASHED_PASSWORD = 'dc9b4c4a06e7dc774e469a16243ff448a02932fd323bb7ecb0ece308d6bb13c964dd7a0b7b99b15dc785004bdb869450f4070a9f83b027af43025149d1bc8f34'
+
+    // hash of 'wrongpassword'
+    def WRONG_SHA_HASHED_PASSWORD = 'bfe2ee0a1bd6ca719f7b6855eaa2ce166e9f06678a662accac9b8f84ba84801d7f76c7c9a05d5207301fee9a6a11d3178eb34c47647835e7f888d3ce6e7935ba'
+
+    def userName = 'userName'
+    def password = 'password'
+    def userId = '618b398c-0110-43f2-95df-d1bc4e7d2b4a'
+
+    SCIMUserProvisioning userProvisioning = Mock()
```
wallner: can you please use `def userprovisioning = Mock(SCIMUserProvisioning)`? Intellij is capable to infer the type and it's more 'Groovy'

src/test/groovy/org/osiam/resources/provisioning/SCIMUserProvisioningBeanSpec.groovy
```diff
@@ -31,15 +31,16 @@ import org.osiam.storage.dao.UserDao
 import org.osiam.storage.entities.MetaEntity
 import org.osiam.storage.entities.UserEntity
 import org.springframework.security.authentication.encoding.PasswordEncoder
+import org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder
 import spock.lang.Specification
 
 class SCIMUserProvisioningBeanSpec extends Specification {
 
-    def passwordEncoder = Mock(PasswordEncoder)
+    BCryptPasswordEncoder bCryptPasswordEncoder = Mock()
```
wallner: Why did you change it to this? The replaced statement is the spock idiom to mocking.

src/test/groovy/org/osiam/auth/login/internal/InternalAuthenticationProviderSpec.groovy
dacrome: Done

src/test/groovy/org/osiam/resources/provisioning/SCIMUserProvisioningBeanSpec.groovy
dacrome: I didn't know that it's the spock idiom. changed.

src/test/groovy/org/osiam/auth/login/internal/InternalAuthenticationProviderSpec.groovy
```diff
@@ -23,15 +23,96 @@
...
+
+    // hash of 'wrongpassword'
+    def WRONG_SHA_HASHED_PASSWORD = 'bfe2ee0a1bd6ca719f7b6855eaa2ce166e9f06678a662accac9b8f84ba84801d7f76c7c9a05d5207301fee9a6a11d3178eb34c47647835e7f888d3ce6e7935ba'
+
+    def userName = 'userName'
+    def password = 'password'
+    def userId = '618b398c-0110-43f2-95df-d1bc4e7d2b4a'
+
+    def userProvisioning = Mock(SCIMUserProvisioning)
+    BCryptPasswordEncoder bCryptPasswordEncoder = new BCryptPasswordEncoder(13)
```
wallner: Please `def` this one too `def bCryptPasswordEncoder = new BCryptPasswordEncoder(13)`

src/test/groovy/org/osiam/auth/login/internal/InternalAuthenticationProviderSpec.groovy
```diff
@@ -23,15 +23,96 @@
...
+
+    @Shared
+    ShaPasswordEncoder shaPasswordEncoder = new ShaPasswordEncoder(512)
```
wallner: Please `def` this one too. `shaPasswordEncoder = new ShaPasswordEncoder(512)`

src/test/groovy/org/osiam/auth/login/internal/InternalAuthenticationProviderSpec.groovy
dacrome: done

src/test/groovy/org/osiam/auth/login/internal/InternalAuthenticationProviderSpec.groovy
dacrome: done

src/test/groovy/org/osiam/resources/provisioning/update/UserUpdaterSpec.groovy
```diff
@@ -48,7 +49,7 @@ class UserUpdaterSpec extends Specification {
     X509CertificateUpdater x509CertificateUpdater = Mock()
     AddressUpdater addressUpdater = Mock()
     UserDao userDao = Mock()
-    PasswordEncoder passwordEncoder = Mock()
+    BCryptPasswordEncoder bCryptPasswordEncoder = Mock()
```
wallner: It's still the wrong way round. And please correct the other variables too.

dacrome: done

dacrome:
I changed the description https://github.com/osiam/osiam/pull/53#issue-113182274

wallner:
Thank you for the work!

PR #91, opened 2015-12-15
Remove unnecessary SSLRequestLoggingFilter

### LINKED ISSUE #74: Remove SSLRequestLoggingFilter
Moved from https://github.com/osiam/resource-server/issues/79

As discussed in #12, we want to get rid of the annoying `SSLRequestLoggingFilter`.

wallner:
Remove the SSLRequestLoggingFilter that does fill the log with lines
requesting, that the SSL should be used. It's not OSIAM's business to
evaluate if the requests to it get encrypted with SSL or not. It might
also drown important log information. Closes #74


tkrille:
Thanks alot!

PR #97, opened 2015-12-27
Update links to old changelogs

tkrille:
Point links to changelog in respective legacy repo.
Makes it much easier to keep them up to date.
Remove old changelog files as part of this update.


wallner:
That's a good idea! Thank you!

PR #100, opened 2016-01-01
Never return null as ClientDetails

tkrille:
The Javadoc of `ClientDetailsService#loadClientByClientId` states that it should never return null [1].
Change the current implementation to fulfill this contract. Instead a `NoSuchClientException` will be thrown, which will be consistently handled by the framework.

[1] https://github.com/spring-projects/spring-security-oauth/blob/2.0.8.RELEASE/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/ClientDetailsService.java#L28


wallner:
Thank you!

PR #101, opened 2016-01-01
Refactor configuration of client credentials grant

tkrille:
Use `spring-security-oauth2`'s ability to add the `ClientCredentialsTokenEndpointFilter` by itself in `OAuth2AuthorizationServerConfig`, therefore remove the addition of the `ClientCredentialsTokenEndpointFilter` in `OAuth2ClientCredentialsSecurity`.

Rename `OAuth2ClientCredentialsSecurity` to `FacebookClientCredentialsSecurity`, because now it only handles the client credentials grant that mimics Facebook's OAuth 2 procces. Set the basic security defaults in `FacebookClientCredentialsSecurity` as they haven't been there before. Set an @Order of 1 to `FacebookClientCredentialsSecurity`, because it has to consulted before the REST security filter chain (3) but after the authorization server security filter chain (0).


wallner:
Thank you!

PR #104, opened 2016-01-03
Re-format the `HttpSecurity` builder calls

tkrille:
They easier to comprehend this way.


wallner:
It's easier to read now, you are right.

Issue #40, opened 2015-10-01
Code-Style for Logger


fwilhe:
Based on osiam/resource-server#80, we had an informal discussion about the naming for loggers.

The discussed options are:
- `LOGGER` (This is the current state in the OSIAM codebase, and IDEs will suggest such names for `static final` variables, because this is how constants are declared in java)
- `LOG`
- `logger`

For comparison, consider how the calls would look like:

``` java
// 1
LOGGER.warn("foo");
// 2
LOG.warn("foo");
// 3
logger.warn("foo");
```

(I did not hear another option until now, feel free to add, if you have one in mind)

I personally would prefer `logger` because although the field is `static final`, it is not a "constant".

This issue contains two tasks:
- Make an decision on how we want to have the style
- (If we decide to not use `LOGGER`): Make the change in the whole OSIAM codebase.

wallner:
My favorite is 3.

tkrille:
``` java
// 4
log.warn("foo");
```

My favorite is 4 :)

dacrome:
My favorite is 4

fwilhe:
Since @osiam-ci is not permitted to vote (I'm looking at you, @dacrome ;-), we have a tie.

I'm not so sure. I don't like such abbreviations.

`log` can mean so many things. Is it a logarithm? Well, I admit that this particular mistake is really not likely provided the context, but still I prefer to remove the mental overhead that the abbreviation introduces.

How do we proceed? Is anyone willing to switch and break the tie?

tkrille:
Should we flip a coin?

dacrome:
alea iacta est: `logger`

PR #108, opened 2016-01-09
Allow colons (:) as field separators for URNs

### LINKED ISSUE #94: Unable to filter resource by extension field
Hi, 

When querying a user to the OSIAM resource server with a multi-filter including an extension field, the server responds with this error:

`Could not deserialize the error response for the HTTP status 'Conflict'. Original response: {"error_code":"CONFLICT","description":"Filtering not possible: 'urn:com.mycompany:scim:extensions:1.0:client' not available"}`

The extension is valid and user has a defined value for the `client` field. 

Is this a bug or a SCIM feature not yet implemented?

wallner:
The scim specification uses colons (:) to separate fields from the rest
of an URN. OSIAM, until now used periods (.) this commit fixes that.
Using periods to separate fields from URNs is still possible, but a
warning is generated informing the user that this is not SCIM conform.

This fixes #94

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,13 @@
     clients and then add the new scopes `ADMIN` and `ME`. See the
     [migration notes] (docs/Migration.md#from-22-to-30) for further details.
 
+- Allow colons (:) as field separators for URNs of extensions, since this is
+  what the SCIM specification defines. Using periods (.) is still possible,
+  but will log a warning message. 
+
+- Fields of the core schemas for user and group can be fully qualified, i.e.
+  `filter=urn:ietf:params:scim:schemas:core:2.0:User:userName sw "J"` 
+
 - All invalid search queries now respond with a `400 BAD REQUEST` instead of
   `409 CONFLICT` status code.
 - Respond with `401 UNAUTHORIZED` when revoking or validating an access token
--- a/src/main/java/org/osiam/storage/query/FilterExpression.java
+++ b/src/main/java/org/osiam/storage/query/FilterExpression.java
@@ -23,24 +23,19 @@
 
 package org.osiam.storage.query;
 
-/**
- * Representation of scim search expression.
- */
-public class ScimExpression {
+import org.osiam.storage.entities.ResourceEntity;
 
-    private final String field;
+public abstract class FilterExpression<T extends ResourceEntity> {
+
     private final FilterConstraint constraint;
     private final String value;
 
-    public ScimExpression(String field, FilterConstraint constraint, String value) {
-        this.field = field;
+    public FilterExpression(FilterConstraint constraint, String value) {
         this.constraint = constraint;
         this.value = value;
     }
 
-    public String getField() {
-        return field;
-    }
+    public abstract Field<T> getField();
 
     public FilterConstraint getConstraint() {
         return constraint;
@@ -48,5 +43,16 @@ public class ScimExpression {
 
     public String getValue() {
         return value;
+    }
+
+    public interface Field<T extends ResourceEntity> {
+
+        String getUrn();
+
+        String getName();
+
+        QueryField<T> getQueryField();
+
+        boolean isExtension();
     }
 }
--- a/src/main/java/org/osiam/storage/query/FilterParser.java
+++ b/src/main/java/org/osiam/storage/query/FilterParser.java
@@ -23,14 +23,14 @@
 
 package org.osiam.storage.query;
 
+import org.antlr.v4.runtime.tree.ParseTree;
+import org.osiam.storage.entities.ResourceEntity;
+
 import javax.persistence.EntityManager;
 import javax.persistence.PersistenceContext;
 import javax.persistence.criteria.Expression;
 import javax.persistence.criteria.Predicate;
 import javax.persistence.criteria.Root;
-
-import org.antlr.v4.runtime.tree.ParseTree;
-import org.osiam.storage.entities.ResourceEntity;
 
 public abstract class FilterParser<T extends ResourceEntity> {
 
@@ -53,8 +53,10 @@ public abstract class FilterParser<T extends ResourceEntity> {
         return filterField.createSortByField(root, entityManager.getCriteriaBuilder());
     }
 
+    public abstract FilterExpression<T> createFilterExpression(String field, FilterConstraint constraint, String value);
+
     protected abstract QueryField<T> getFilterField(String sortBy);
 
-    protected abstract FilterChain<T> createFilterChain(ScimExpression filter);
+    protected abstract FilterChain<T> createFilterChain(FilterExpression<T> filter);
 
 }
--- a/src/main/java/org/osiam/storage/query/GroupFilterExpression.java
+++ b/src/main/java/org/osiam/storage/query/GroupFilterExpression.java
@@ -1,0 +1,67 @@
+package org.osiam.storage.query;
+
+import org.osiam.resources.scim.Group;
+import org.osiam.storage.entities.GroupEntity;
+
+import java.util.Locale;
+
+public class GroupFilterExpression extends FilterExpression<GroupEntity> {
+
+    private final Field field;
+
+    public GroupFilterExpression(String field, FilterConstraint constraint, String value) {
+        super(constraint, value);
+        this.field = new Field(field);
+    }
+
+    @Override
+    public Field getField() {
+        return field;
+    }
+
+    public static class Field implements FilterExpression.Field<GroupEntity> {
+        private String name;
+        private QueryField<GroupEntity> queryField;
+
+        public Field(String field) {
+
+            if (field.toLowerCase().startsWith(Group.SCHEMA.toLowerCase())) {
+                if (field.charAt(Group.SCHEMA.length()) == '.') {
+                    throw new IllegalArgumentException(String.format("Period (.) is not a valid field separator: %s", field));
+                }
+
+                name = field.substring(Group.SCHEMA.length() + 1).toLowerCase(Locale.ENGLISH);
+
+            } else {
+                name = field.toLowerCase(Locale.ENGLISH);
+            }
+            queryField = GroupQueryField.fromString(name);
+
+            if (queryField == null) {
+                throw new IllegalArgumentException(String.format("Unable to parse %s, extensions not supported for groups",
+                        field));
+            }
+        }
+
+        @Override
+        public String getUrn() {
+            // We are not supporting Group extensions at the moment.
+            return Group.SCHEMA;
+        }
+
+        @Override
+        public String getName() {
+            return name;
+        }
+
+        @Override
+        public QueryField<GroupEntity> getQueryField() {
+            return queryField;
+        }
+
+        @Override
+        public boolean isExtension() {
+            return false;
+        }
+    }
+}
--- a/src/main/java/org/osiam/storage/query/GroupSimpleFilterChain.java
+++ b/src/main/java/org/osiam/storage/query/GroupSimpleFilterChain.java
@@ -23,35 +23,27 @@
 
 package org.osiam.storage.query;
 
-import java.util.Locale;
+import org.osiam.storage.entities.GroupEntity;
 
 import javax.persistence.criteria.CriteriaBuilder;
 import javax.persistence.criteria.Predicate;
 import javax.persistence.criteria.Root;
 
-import org.osiam.storage.entities.GroupEntity;
-
 public class GroupSimpleFilterChain implements FilterChain<GroupEntity> {
 
-    private final ScimExpression scimExpression;
+    private final FilterExpression<GroupEntity> filterExpression;
 
-    private final GroupQueryField filterField;
+    private final QueryField<GroupEntity> filterField;
     private final CriteriaBuilder criteriaBuilder;
 
-    public GroupSimpleFilterChain(CriteriaBuilder criteriaBuilder, ScimExpression scimExpression) {
+    public GroupSimpleFilterChain(CriteriaBuilder criteriaBuilder, FilterExpression<GroupEntity> filterExpression) {
         this.criteriaBuilder = criteriaBuilder;
-        this.scimExpression = scimExpression;
-        filterField = GroupQueryField.fromString(scimExpression.getField().toLowerCase(Locale.ENGLISH));
+        this.filterExpression = filterExpression;
+        filterField = filterExpression.getField().getQueryField();
     }
 
     @Override
     public Predicate createPredicateAndJoin(Root<GroupEntity> root) {
-        if (filterField == null) {
-            throw new IllegalArgumentException(String.format("Unable to filter. Field '%s' not available.",
-                    scimExpression.getField()));
-        }
-
-        return filterField.addFilter(root, scimExpression.getConstraint(), scimExpression.getValue(), criteriaBuilder);
+        return filterField.addFilter(root, filterExpression.getConstraint(), filterExpression.getValue(), criteriaBuilder);
     }
-
 }
--- a/src/main/java/org/osiam/storage/query/UserFilterExpression.java
+++ b/src/main/java/org/osiam/storage/query/UserFilterExpression.java
@@ -1,0 +1,93 @@
+package org.osiam.storage.query;
+
+import org.osiam.resources.scim.User;
+import org.osiam.storage.entities.UserEntity;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Locale;
+
+public class UserFilterExpression extends FilterExpression<UserEntity> {
+
+    private Field field;
+
+    UserFilterExpression(String field, FilterConstraint constraint, String value) {
+        super(constraint, value);
+        this.field = new Field(field);
+    }
+
+    @Override
+    public UserFilterExpression.Field getField() {
+        return field;
+    }
+
+    public static class Field implements FilterExpression.Field<UserEntity> {
+        private static final Logger logger = LoggerFactory.getLogger(UserFilterExpression.Field.class);
+        private String urn;
+        private String name;
+        private UserQueryField queryField;
+
+        public Field(String field) {
+            if (field.toLowerCase().startsWith(User.SCHEMA.toLowerCase())) {
+                if (field.charAt(User.SCHEMA.length()) == '.') {
+                    throw new IllegalArgumentException(String.format("Period (.) is not a valid field separator: %s", field));
+                }
+                name = field.substring(User.SCHEMA.length() + 1).toLowerCase(Locale.ENGLISH);
+            } else {
+                name = field.toLowerCase(Locale.ENGLISH);
+            }
+
+            // Try to resolve a core schema field.
+            queryField = UserQueryField.fromString(name);
+
+            urn = User.SCHEMA;
+            // we can't recognize a core schema field, it's an extension.
+            if (queryField == null) {
+
+                // lastIndexOf returns -1 if char can not be found in the string
+                int lastIndexOfPeriod = field.lastIndexOf('.');
+                int lastIndexOfColon = field.lastIndexOf(':');
+
+                // the last string part of the string is separated by a period, that is not SCIM conform but has been
+                // supported by OSIAM for a long time so we generate a warning. As soon as we support extended attributes
+                // for extensions this needs to go.
+                if (lastIndexOfPeriod > lastIndexOfColon) {
+                    urn = field.substring(0, lastIndexOfPeriod);
+                    name = field.substring(lastIndexOfPeriod + 1);
+                    logger.warn(
+                            String.format("Period (.) used as field separator in %s. This is not SCIM conform and deprecated!",
+                                    field));
+                } else if (lastIndexOfColon > lastIndexOfPeriod) {
+                    // a colon is separating the last part of this string and the field of the extension from it's URN.
+                    // Just as it's supposed to be.
+                    urn = field.substring(0, lastIndexOfColon);
+                    name = field.substring(lastIndexOfColon + 1);
+                } else {
+                    // Neither colon nor period found in string. It's a user error.
+                    throw new IllegalArgumentException(String.format("Unable to parse field or extension from %s",
+                            field));
+                }
+            }
+        }
+
+        @Override
+        public String getUrn() {
+            return urn;
+        }
+
+        @Override
+        public String getName() {
+            return name;
+        }
+
+        @Override
+        public UserQueryField getQueryField() {
+            return queryField;
+        }
+
+        @Override
+        public boolean isExtension() {
+            return queryField == null;
+        }
+    }
+}
--- a/src/main/java/org/osiam/storage/query/UserFilterParser.java
+++ b/src/main/java/org/osiam/storage/query/UserFilterParser.java
@@ -41,8 +41,8 @@ public class UserFilterParser extends FilterParser<UserEntity> {
     }
 
     @Override
-    protected FilterChain<UserEntity> createFilterChain(ScimExpression filter) {
-        return new UserSimpleFilterChain(entityManager.getCriteriaBuilder(), extensionDao, filter);
+    public FilterExpression<UserEntity> createFilterExpression(String field, FilterConstraint constraint, String value) {
+        return new UserFilterExpression(field, constraint, value);
     }
 
     @Override
@@ -50,4 +50,8 @@ public class UserFilterParser extends FilterParser<UserEntity> {
         return UserQueryField.fromString(sortBy.toLowerCase(Locale.ENGLISH));
     }
 
+    @Override
+    protected UserSimpleFilterChain createFilterChain(FilterExpression<UserEntity> filter) {
+        return new UserSimpleFilterChain(entityManager.getCriteriaBuilder(), extensionDao, filter);
+    }
 }
--- a/src/main/java/org/osiam/storage/query/UserSimpleFilterChain.java
+++ b/src/main/java/org/osiam/storage/query/UserSimpleFilterChain.java
@@ -32,65 +32,50 @@ import org.osiam.storage.entities.UserEntity;
 import javax.persistence.criteria.CriteriaBuilder;
 import javax.persistence.criteria.Predicate;
 import javax.persistence.criteria.Root;
-import java.util.Locale;
 
 public class UserSimpleFilterChain implements FilterChain<UserEntity> {
 
-    private final ScimExpression scimExpression;
+    private final FilterExpression<UserEntity> filterExpression;
 
     private final QueryField<UserEntity> userFilterField;
     private final ExtensionDao extensionDao;
     private final CriteriaBuilder criteriaBuilder;
-    private ExtensionQueryField extensionFilterField;
+    private final ExtensionQueryField extensionFilterField;
 
     public UserSimpleFilterChain(CriteriaBuilder criteriaBuilder, ExtensionDao extensionDao,
-                                 ScimExpression scimExpression) {
+                                 FilterExpression<UserEntity> filterExpression) {
         this.criteriaBuilder = criteriaBuilder;
         this.extensionDao = extensionDao;
-        this.scimExpression = scimExpression;
+        this.filterExpression = filterExpression;
 
-        String field = scimExpression.getField();
-
-        userFilterField = UserQueryField.fromString(field.toLowerCase(Locale.ENGLISH));
-
-        // It's not a known user field, so try to build a extension filter
-        if (userFilterField == null) {
-            extensionFilterField = getExtensionFilterField(field.toLowerCase(Locale.ENGLISH));
-            if (extensionFilterField == null) {
-                throw new IllegalArgumentException(String.format("Filtering not possible: '%s' not available", field));
-            }
+        if (filterExpression.getField().isExtension()) {
+            extensionFilterField = getExtensionFilterField(filterExpression);
+            userFilterField = null;
+        } else {
+            userFilterField = filterExpression.getField().getQueryField();
+            extensionFilterField = null;
         }
     }
 
-    private ExtensionQueryField getExtensionFilterField(String fieldString) {
-        int lastIndexOf = fieldString.lastIndexOf('.');
-        if (lastIndexOf == -1) {
-            return null;
-        }
-
-        String urn = fieldString.substring(0, lastIndexOf);
-        String fieldName = fieldString.substring(lastIndexOf + 1);
+    private ExtensionQueryField getExtensionFilterField(FilterExpression<UserEntity> filterExpression) {
         final ExtensionEntity extension;
         try {
-            extension = extensionDao.getExtensionByUrn(urn, true);
+            extension = extensionDao.getExtensionByUrn(filterExpression.getField().getUrn(), true);
         } catch (OsiamException ex) {
-            return null;
+            throw new IllegalArgumentException(String.format("Filtering not possible. Field '%s' not available.", filterExpression.getField()));
         }
-        final ExtensionFieldEntity fieldEntity = extension.getFieldForName(fieldName, true);
-        return new ExtensionQueryField(urn, fieldEntity);
+        final ExtensionFieldEntity fieldEntity = extension.getFieldForName(filterExpression.getField().getName(), true);
+        return new ExtensionQueryField(filterExpression.getField().getUrn(), fieldEntity);
     }
 
     @Override
     public Predicate createPredicateAndJoin(Root<UserEntity> root) {
         if (userFilterField != null) {
-            return userFilterField.addFilter(root, scimExpression.getConstraint(), scimExpression.getValue(),
-                    criteriaBuilder);
-        } else if (extensionFilterField != null) {
-            return extensionFilterField.addFilter(root, scimExpression.getConstraint(), scimExpression.getValue(),
+            return userFilterField.addFilter(root, filterExpression.getConstraint(), filterExpression.getValue(),
                     criteriaBuilder);
         } else {
-            throw new IllegalArgumentException(String.format("Filtering not possible. Field '%s' not available.",
-                    scimExpression.getField()));
+            return extensionFilterField.addFilter(root, filterExpression.getConstraint(), filterExpression.getValue(),
+                    criteriaBuilder);
         }
     }
 }
--- a/src/test/groovy/org/osiam/storage/query/GroupFilterExpressionSpec.groovy
+++ b/src/test/groovy/org/osiam/storage/query/GroupFilterExpressionSpec.groovy
@@ -1,0 +1,58 @@
+package org.osiam.storage.query
+
+import org.osiam.resources.scim.Group
+import spock.lang.Specification
+import spock.lang.Unroll
+
+class GroupFilterExpressionSpec extends Specification {
+
+    @Unroll
+    def 'QueryField "#queryField" can be resolved'() {
+        given:
+        def expression = new GroupFilterExpression("${queryField}", FilterConstraint.EQUALS, 'irrelevant')
+
+        expect:
+        expression.field.queryField.toString() == queryField
+        expression.field.getUrn() == Group.SCHEMA
+        !expression.field.isExtension()
+
+        where:
+        queryField << GroupQueryField.values().collect { it.toString() }
+    }
+
+    @Unroll
+    def 'Resolving fully qualified query field "#queryField"'() {
+        given:
+        def expression = new GroupFilterExpression("${Group.SCHEMA}:${queryField}", FilterConstraint.EQUALS, 'irrelevant')
+
+        expect:
+        expression.field.queryField.toString() == queryField
+        expression.field.getUrn() == Group.SCHEMA
+        !expression.field.isExtension()
+
+        where:
+        queryField << GroupQueryField.values().collect { it.toString() }
+    }
+
+    @Unroll
+    def 'Separating field "#queryField" by period rasises an exception'() {
+        when:
+        def expression = new GroupFilterExpression("${Group.SCHEMA}.${queryField}", FilterConstraint.EQUALS, 'irrelevant')
+
+        then:
+        thrown(IllegalArgumentException)
+
+        where:
+        queryField << GroupQueryField.values().collect { it.toString() }
+    }
+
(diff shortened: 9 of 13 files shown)


src/main/java/org/osiam/storage/query/ScimExpression.java
```diff
@@ -49,4 +55,57 @@ public FilterConstraint getConstraint() {
...
+
+        public UserQueryField getQueryField() {
+            return queryField;
+        }
+
+        public boolean isExtension() {
+            return queryField == null;
+        }
+
+
```
tkrille, line 55: Please remove 1 empty line.

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@
     clients and then add the new scopes `ADMIN` and `ME`. See the
     [migration notes] (docs/Migration.md#from-22-to-30) for further details.
 
+- Allow colons (:) as field separators for URNs of extensions, since this is
+  what the SCIM specification defines. Using periods (.) is still possible,
+  but will log a warning message. 
+
```
tkrille, line 20: Most of the changelog entries take the form:

```
- <Some brief summary>

    <An optional, more descriptive text>
```

Don't get me wrong, I also like your style and I don't want to force usage of the old style or any style whatsoever. I'm just wondering whether we should agree on one style. As this is a fresh new changelog, anything is possible.

tkrille, line 20: And isn't there an entry missing? Isn't it now possible to also send the full-qualified field name, if it is a core field, i.e. sending:

```
urn:ietf:params:scim:schemas:core:2.0:User:userName eq 'John'
```

Which will work now and must also work according to the spec.

src/main/java/org/osiam/storage/query/ScimExpression.java
```diff
@@ -23,22 +23,28 @@
...
+import java.util.Locale;
+
 /**
  * Representation of scim search expression.
  */
 public class ScimExpression {
 
-    private final String field;
+    private static final Logger LOG = LoggerFactory.getLogger(ScimExpression.class);
```
tkrille: I think [we once agreed](https://github.com/osiam/osiam/issues/40) on how we name loggers.

tkrille: And shouldn't the logger belong the `Field` class? It's only used there.

CHANGELOG.md
tkrille, line 20: Alright, let's leave the changelog entry as it is. But nevertheless, I think my [previous comment](https://github.com/osiam/osiam/pull/108#discussion_r49280862) is still valid.

wallner:
All of your comments are right of course and the changes already happened, but a lot changed in my working copy. I'm locking this PR for the moment.

wallner:
The one thing I'm not happy with is when the lower casing of filters and fields happens. At the moment we are doing that when we are building the `Field<ResourceEntity>` i think we need to do it way earlier. My suggestion would be to do it right in the controller. What do you think?

tkrille:
Can I see some ITs for this, please?

tkrille:
Regarding your concerns: I'd say we should even lowercase the values as late as possible.

src/main/java/org/osiam/storage/query/GroupFilterExpression.java
```diff
@@ -0,0 +1,70 @@
...
+        return field;
+    }
+
+    public static class Field implements FilterExpression.Field<GroupEntity> {
+        private String fieldName;
+        private QueryField queryField;
+
+        public Field(String field) {
+
+
```
tkrille, line 32: Please remove this empty line.

src/main/java/org/osiam/storage/query/GroupFilterExpression.java
```diff
@@ -0,0 +1,70 @@
...
+        public QueryField getQueryField() {
+            return queryField;
+        }
+
+        @Override
+        public boolean isExtension() {
+            return false;
+        }
+
+        private void splitString(String urnAndField, int lastIndexOf) {
```
tkrille: Unused method

src/main/java/org/osiam/storage/query/FilterExpression.java
```diff
@@ -23,24 +23,26 @@
 
 package org.osiam.storage.query;
 
+import org.osiam.storage.entities.ResourceEntity;
+import org.slf4j.Logger;
```
tkrille: The last 3 imports are unused.

src/main/java/org/osiam/storage/query/FilterExpression.java
```diff
@@ -23,24 +23,26 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.Locale;
+
 /**
  * Representation of scim search expression.
```
tkrille: Could you improve the javadoc? It's at least a "SCIM search filter expression" :)

src/main/java/org/osiam/storage/query/FilterExpression.java
```diff
@@ -49,4 +51,15 @@ public FilterConstraint getConstraint() {
     public String getValue() {
         return value;
     }
-}
\ No newline at end of file
+
+    public static interface Field<T> {
```
tkrille: Remove `static`

tkrille: Type parameter must be `T extends ResourceEntity`.

src/main/java/org/osiam/storage/query/FilterExpression.java
```diff
@@ -49,4 +51,15 @@ public FilterConstraint getConstraint() {
...
+
+        String getUrn();
+
+        String getFieldName();
+
+        QueryField getQueryField();
```
tkrille: Missing type parameter `<T>`

src/main/java/org/osiam/storage/query/FilterExpression.java
```diff
@@ -23,24 +23,26 @@
...
-        this.field = field;
+    public FilterExpression(FilterConstraint constraint, String value) {
         this.constraint = constraint;
         this.value = value;
     }
 
-    public String getField() {
-        return field;
-    }
+    abstract public Field<T> getField();
```
tkrille: I think it's `public abstract`

src/main/java/org/osiam/storage/query/FilterParser.java
```diff
@@ -53,8 +53,10 @@ public Predicate createPredicateAndJoin(ParseTree filterTree, Root<T> root) {
         return filterField.createSortByField(root, entityManager.getCriteriaBuilder());
     }
 
+    public abstract FilterExpression<T> buildFilterExpression(String field, FilterConstraint constraint, String value);
```
tkrille: Please rename to `create`, like the other methods.

src/main/java/org/osiam/storage/query/FilterExpression.java
```diff
@@ -23,24 +23,26 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Locale;
+
 /**
  * Representation of scim search expression.
  */
-public class ScimExpression {
+public abstract class FilterExpression<T extends ResourceEntity> {
```
tkrille, line 28: You made this class generic, but there are many places that don't have been updated to reflect this. Please review the use sites of the class.

src/main/java/org/osiam/storage/query/GroupFilterExpression.java
```diff
@@ -0,0 +1,70 @@
...
+    public static class Field implements FilterExpression.Field<GroupEntity> {
+        private String fieldName;
+        private QueryField queryField;
+
+        public Field(String field) {
+
+
+            if (field.startsWith(Group.SCHEMA)) {
+                fieldName = field.substring(Group.SCHEMA.length() + 1).toLowerCase(Locale.ENGLISH);
+
```
tkrille, line 34: Please remove this empty line.

src/main/java/org/osiam/storage/query/FilterExpression.java
```diff
@@ -49,4 +51,15 @@ public FilterConstraint getConstraint() {
...
         return value;
     }
-}
\ No newline at end of file
+
+    public static interface Field<T> {
+
+        String getUrn();
+
+        String getFieldName();
```
tkrille: Why not `getName`?

src/main/java/org/osiam/storage/query/GroupSimpleFilterChain.java
```diff
@@ -33,25 +31,25 @@
...
     private final GroupQueryField filterField;
     private final CriteriaBuilder criteriaBuilder;
 
-    public GroupSimpleFilterChain(CriteriaBuilder criteriaBuilder, ScimExpression scimExpression) {
+    public GroupSimpleFilterChain(CriteriaBuilder criteriaBuilder, FilterExpression filterExpression) {
         this.criteriaBuilder = criteriaBuilder;
-        this.scimExpression = scimExpression;
-        filterField = GroupQueryField.fromString(scimExpression.getField().toLowerCase(Locale.ENGLISH));
+        this.filterExpression = filterExpression;
+        filterField = GroupQueryField.fromString(filterExpression.getField().getFieldName());
```
tkrille: The filter field can be directly retrieved: `filterExpression.getField().getQueryField()`.

src/main/java/org/osiam/storage/query/GroupSimpleFilterChain.java
```diff
@@ -33,25 +31,25 @@
...
 public class GroupSimpleFilterChain implements FilterChain<GroupEntity> {
 
-    private final ScimExpression scimExpression;
+    private final FilterExpression filterExpression;
 
     private final GroupQueryField filterField;
     private final CriteriaBuilder criteriaBuilder;
 
-    public GroupSimpleFilterChain(CriteriaBuilder criteriaBuilder, ScimExpression scimExpression) {
+    public GroupSimpleFilterChain(CriteriaBuilder criteriaBuilder, FilterExpression filterExpression) {
```
tkrille: Shouldn't this be `GroupFilterExpression`?

src/main/java/org/osiam/storage/query/GroupSimpleFilterChain.java
```diff
@@ -33,25 +31,25 @@
...
         this.criteriaBuilder = criteriaBuilder;
-        this.scimExpression = scimExpression;
-        filterField = GroupQueryField.fromString(scimExpression.getField().toLowerCase(Locale.ENGLISH));
+        this.filterExpression = filterExpression;
+        filterField = GroupQueryField.fromString(filterExpression.getField().getFieldName());
     }
 
     @Override
     public Predicate createPredicateAndJoin(Root<GroupEntity> root) {
         if (filterField == null) {
```
tkrille: You can remove this guard clause, if you follow the [previous comment](https://github.com/osiam/osiam/pull/108/files#r49906584).

src/main/java/org/osiam/storage/query/UserFilterExpression.java
```diff
@@ -0,0 +1,87 @@
...
+import org.osiam.storage.entities.UserEntity;
+import org.slf4j.LoggerFactory;
+
+import java.util.Locale;
+
+public class UserFilterExpression extends FilterExpression<UserEntity> {
+
+    private Field field;
+
+
```
tkrille: Remove this empty line

src/main/java/org/osiam/storage/query/UserFilterExpression.java
```diff
@@ -0,0 +1,87 @@
...
+        this.field = new Field(field);
+    }
+
+    @Override
+    public UserFilterExpression.Field getField() {
+        return field;
+    }
+
+    public static class Field implements FilterExpression.Field<UserEntity> {
+        private static final org.slf4j.Logger LOG = LoggerFactory.getLogger(UserFilterExpression.Field.class);
```
tkrille: Really? The whole FQCN? :)

src/main/java/org/osiam/storage/query/UserFilterExpression.java
```diff
@@ -0,0 +1,87 @@
...
+        private String urn;
+        private String fieldName;
+        private QueryField queryField;
+
+        public Field(String field) {
+            if (field.startsWith(User.SCHEMA)) {
+                fieldName = field.substring(User.SCHEMA.length() + 1).toLowerCase(Locale.ENGLISH);
+
```
tkrille: Remove this empty line

src/main/java/org/osiam/storage/query/GroupFilterExpression.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    public static class Field implements FilterExpression.Field<GroupEntity> {
+        private String fieldName;
+        private QueryField queryField;
+
+        public Field(String field) {
+
+
+            if (field.startsWith(Group.SCHEMA)) {
```
tkrille: I think URNs must be case-insensitive.

src/main/java/org/osiam/storage/query/UserFilterExpression.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    public static class Field implements FilterExpression.Field<UserEntity> {
+        private static final org.slf4j.Logger LOG = LoggerFactory.getLogger(UserFilterExpression.Field.class);
+        private String urn;
+        private String fieldName;
+        private QueryField queryField;
+
+        public Field(String field) {
+            if (field.startsWith(User.SCHEMA)) {
```
tkrille: I think URNs must be case-insensitive.

src/main/java/org/osiam/storage/query/UserFilterExpression.java
```diff
@@ -0,0 +1,87 @@
...
+
+            queryField = UserQueryField.fromString(fieldName);
+
+            if (queryField == null) {
+
+                int lastIndexOfPeriod = field.lastIndexOf('.');
+                int lastIndexOfColon = field.lastIndexOf(':');
+
+                if (lastIndexOfPeriod > lastIndexOfColon) {
+                    // separated by colon, log a warning!
```
tkrille: You mean "period", don't you?

src/main/java/org/osiam/storage/query/UserFilterExpression.java
```diff
@@ -0,0 +1,87 @@
...
+                    splitString(field, lastIndexOfPeriod);
+                    LOG.warn(
+                            String.format("Period (.) used as field separator in %s. " +
```
tkrille: Could you join this line with the one before?

src/main/java/org/osiam/storage/query/UserFilterExpression.java
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        @Override
+        public boolean isExtension() {
+            return queryField == null;
+        }
+
+        private void splitString(String urnAndField, int lastIndexOf) {
+            urn = urnAndField.substring(0, lastIndexOf);
+            fieldName = urnAndField.substring(lastIndexOf + 1);
```
tkrille: I think you have to lower case this.

src/main/java/org/osiam/storage/query/UserFilterExpression.java
```diff
@@ -0,0 +1,87 @@
...
+        public QueryField getQueryField() {
+            return queryField;
+        }
+
+        @Override
+        public boolean isExtension() {
+            return queryField == null;
+        }
+
+        private void splitString(String urnAndField, int lastIndexOf) {
```
tkrille: I don't like this extracted method. It's not obvious that it has a side effect and therefore the name is misleading. It's also not obvious what the process in the constructor does. I'd suggest to inline this method.

src/main/java/org/osiam/storage/query/UserFilterExpression.java
```diff
@@ -0,0 +1,87 @@
...
+                            String.format("Period (.) used as field separator in %s. " +
+                                    "This is not SCIM conform and deprecated!", field));
+                } else if (lastIndexOfColon > lastIndexOfPeriod) {
+                    splitString(field, lastIndexOfColon);
+                } else {
+                    throw new IllegalArgumentException(String.format("Unable to parse field or extension from %s",
+                            field));
+                }
+            } else {
+                urn = User.SCHEMA;
```
tkrille: You can move this up, before the current `if` statement. If `queryField` is `null` then `urn` will be overwritten or an exception will be thrown.

src/main/java/org/osiam/storage/query/UserFilterParser.java
```diff
@@ -41,11 +41,16 @@ public UserFilterParser(ExtensionDao extensionDao) {
     }
 
     @Override
-    protected FilterChain<UserEntity> createFilterChain(ScimExpression filter) {
+    protected FilterChain<UserEntity> createFilterChain(FilterExpression filter) {
         return new UserSimpleFilterChain(entityManager.getCriteriaBuilder(), extensionDao, filter);
     }
 
     @Override
+    public FilterExpression<UserEntity> buildFilterExpression(String field, FilterConstraint constraint, String value) {
```
tkrille: Please rename to `create`, like the other methods.

src/main/java/org/osiam/storage/query/UserSimpleFilterChain.java
```diff
@@ -44,53 +43,42 @@
...
-        if (userFilterField == null) {
-            extensionFilterField = getExtensionFilterField(field.toLowerCase(Locale.ENGLISH));
-            if (extensionFilterField == null) {
-                throw new IllegalArgumentException(String.format("Filtering not possible: '%s' not available", field));
-            }
+        if (filterExpression.getField().isExtension()) {
+            extensionFilterField = getExtensionFilterField(filterExpression);
+            userFilterField = null;
+        } else {
+            userFilterField = filterExpression.getField().getQueryField();
```
tkrille, line 55: > Unchecked assignment: 'org.osiam.storage.query.QueryField' to 'org.osiam.storage.query.QueryField<org.osiam.storage.entities.UserEntity>'. Reason: 'filterExpression.getField()' has raw type, so result of getQueryField is erased.

src/main/java/org/osiam/storage/query/UserSimpleFilterChain.java
```diff
@@ -44,53 +43,42 @@
...
 
     public UserSimpleFilterChain(CriteriaBuilder criteriaBuilder, ExtensionDao extensionDao,
-                                 ScimExpression scimExpression) {
+                                 FilterExpression filterExpression) {
         this.criteriaBuilder = criteriaBuilder;
         this.extensionDao = extensionDao;
-        this.scimExpression = scimExpression;
+        this.filterExpression = filterExpression;
 
-        String field = scimExpression.getField();
```
tkrille: Remove this empty line

src/main/java/org/osiam/storage/query/UserSimpleFilterChain.java
```diff
@@ -44,53 +43,42 @@
...
-
-        // It's not a known user field, so try to build a extension filter
-        if (userFilterField == null) {
-            extensionFilterField = getExtensionFilterField(field.toLowerCase(Locale.ENGLISH));
-            if (extensionFilterField == null) {
-                throw new IllegalArgumentException(String.format("Filtering not possible: '%s' not available", field));
-            }
+        if (filterExpression.getField().isExtension()) {
+            extensionFilterField = getExtensionFilterField(filterExpression);
+            userFilterField = null;
```
tkrille, line 53: I don't think you have to explicitly set this to null.

src/main/java/org/osiam/storage/query/UserSimpleFilterChain.java
```diff
@@ -44,53 +43,42 @@
...
 
-        String field = scimExpression.getField();
 
-        userFilterField = UserQueryField.fromString(field.toLowerCase(Locale.ENGLISH));
-
-        // It's not a known user field, so try to build a extension filter
-        if (userFilterField == null) {
-            extensionFilterField = getExtensionFilterField(field.toLowerCase(Locale.ENGLISH));
-            if (extensionFilterField == null) {
-                throw new IllegalArgumentException(String.format("Filtering not possible: '%s' not available", field));
```
tkrille, line 60: You moved this error handling into `createPredicateAndJoin`. I think failing fast here is a more viable approach. You have to consider, that `createPredicateAndJoin` gets called very late, at a point when all `FilterChain`s and `FilterExpression`s have already been created. So multiple database accesses, to get the extensions fields, might already have happened. But all this would be thrown away, if there is a wrong field in the filter expression.

src/test/groovy/org/osiam/storage/query/GroupFilterExpressionSpec.groovy
```diff
@@ -0,0 +1,60 @@
...
+        expression.field.queryField.toString() == queryField
+        expression.field.getUrn() == Group.SCHEMA
+        !expression.field.isExtension()
+
+        where:
+        queryField << GroupQueryField.values().collect { it.toString() }
+    }
+
+    @Unroll
+    def 'Resolving field "#queryField" separated by colon is possible'() {
```
tkrille: Again, I think you mean "period" :)

tkrille: Wait a moment… Why should this be working? It didn't work before, and shouldn't work according to the SCIM spec.

src/test/groovy/org/osiam/storage/query/UserFilterExpressionSpec.groovy
```diff
@@ -0,0 +1,81 @@
...
+        ((AppenderAttachable<ILoggingEvent>) rootLogger).addAppender(appender)
+
+        when:
+        def expression = new UserFilterExpression("urn.field", FilterConstraint.EQUALS, "irrelevant")
+
+        then:
+        expression.field.fieldName == 'field'
+        expression.field.urn == 'urn'
+        1 * appender.doAppend({
+            it.message == 'Period (.) used as field separator in urn.field. This is not SCIM conform and deprecated!'
```
tkrille, line 88: Is relying on the exact message really a good thing?

src/test/groovy/org/osiam/storage/query/GroupFilterExpressionSpec.groovy
tkrille: Please, also remove support for `<Core URN>.<Core Field>` from the `UserFilterExpression`.

tkrille:
Hi @wallner,

I reviewed your solution and it looks good to me. But there are several issues with the actually changed code. Please amend your PR and I will do a second brief review and finally merge this.

src/main/java/org/osiam/storage/query/GroupFilterExpression.java
wallner: Oops. How did I miss that.

src/main/java/org/osiam/storage/query/FilterExpression.java
wallner: Sure enough. Can you suggest changes?

tkrille: I already did:

> It's at least a "SCIM search filter expression"

:)

tkrille: But seriously, when in doubt, just remove the javadoc. I mean, does it really explain something that needs to be explained?

wallner: Well, yes. I did that.

src/main/java/org/osiam/storage/query/UserFilterExpression.java
wallner: I'm happy to, but then the line is way wider than 120 characters.

src/main/java/org/osiam/storage/query/UserFilterExpression.java
wallner: Why? This method only operates on the string lengths. Case should not matter here. But maybe I'm missing something.

src/main/java/org/osiam/storage/query/UserFilterExpression.java
wallner: It's a leftover from an earlier version. I will inline it.

src/main/java/org/osiam/storage/query/UserSimpleFilterChain.java
wallner, line 55: It's not anymore

src/test/groovy/org/osiam/storage/query/UserFilterExpressionSpec.groovy
wallner, line 88: Yes, as long as we don't use localized log mesages,I think it's a good thing to verify the message here. Let me optimize it a little.

src/test/groovy/org/osiam/storage/query/GroupFilterExpressionSpec.groovy
wallner: I don't know. I think we should keep it and not differentiate between core and extension schemas. Separating wit period is going to go away and logs a warning message. Before this the querying with fully qualified core schemas didn't work at all.

src/main/java/org/osiam/storage/query/UserSimpleFilterChain.java
wallner, line 53: If we want to keep the field final we do.

src/main/java/org/osiam/storage/query/UserSimpleFilterChain.java
wallner, line 60: I agree

src/main/java/org/osiam/storage/query/UserFilterParser.java
```diff
@@ -50,4 +50,8 @@ public UserFilterParser(ExtensionDao extensionDao) {
         return UserQueryField.fromString(sortBy.toLowerCase(Locale.ENGLISH));
     }
 
+    @Override
+    protected UserSimpleFilterChain createFilterChain(FilterExpression<UserEntity> filter) {
+        return new UserSimpleFilterChain(entityManager.getCriteriaBuilder(), extensionDao, (UserFilterExpression) filter);
```
wallner: Ok, this really stomps me. I don't see why have to use the cast here, but if I change the method signature to the concrete `UserFilterExpression` it doesn't work. Any hints?

src/main/java/org/osiam/storage/query/UserSimpleFilterChain.java
tkrille, line 53: Oh, I see. But mustn't `extensionFilterField` be final too then?

src/main/java/org/osiam/storage/query/UserFilterParser.java
tkrille: Because it's a different method then. The abstract method needs any `FilterExpression<UserEntity>` not just instances of `UserFilterExpression`. I'll take a look at this.

tkrille: It works, if you change:
1. [this line](https://github.com/osiam/osiam/pull/108/files#diff-a67f658e1bef24b05cb4e486780572dcR38) to `private final FilterExpression<UserEntity> filterExpression;`
2. [this line](https://github.com/osiam/osiam/pull/108/files#diff-a67f658e1bef24b05cb4e486780572dcR40) to `private final QueryField<UserEntity> userFilterField;`
3. [this line](https://github.com/osiam/osiam/pull/108/files#diff-a67f658e1bef24b05cb4e486780572dcR46) to `FilterExpression<UserEntity> filterExpression`
4. [this line](https://github.com/osiam/osiam/pull/108/files#diff-a67f658e1bef24b05cb4e486780572dcR59) to `private ExtensionQueryField getExtensionFilterField(FilterExpression<UserEntity> filterExpression) {`

You don't need the concrete classes there anymore. All that counts, is that the given instances implement the interfaces, because you've hidden the actual classes in factories, like the sub-classes of `FilterExpression<T>`.

src/main/java/org/osiam/storage/query/GroupFilterExpression.java
```diff
@@ -0,0 +1,63 @@
...
+        return field;
+    }
+
+    public static class Field implements FilterExpression.Field<GroupEntity> {
+        private String fieldName;
+        private GroupQueryField queryField;
+
+        public Field(String field) {
+
+            if (field.startsWith(Group.SCHEMA)) {
```
tkrille: This still bothers me. I think we should be case-insensitive here. We are at least case-insensitive when checking the URN of an extension. I think this is because of the text in the [SCIM RFC, p. 17](http://tools.ietf.org/html/rfc7644#page-17):

>    Attribute names and attribute operators used in filters are case
>    insensitive.  For example, the following two expressions will
>    evaluate to the same logical value:
> 
>    filter=userName Eq "john"
> 
>    filter=Username eq "john"

But I don't know, whether this applies just to the attribute name or to the URN too. What do you think?

src/main/java/org/osiam/storage/query/UserFilterExpression.java
```diff
@@ -0,0 +1,81 @@
...
+
+    public static class Field implements FilterExpression.Field<UserEntity> {
+        private static final Logger LOG = LoggerFactory.getLogger(UserFilterExpression.Field.class);
+        public static final String LOG_MESSAGE = "Period (.) used as field separator in %s. This is not SCIM conform and deprecated!";
+        private String urn;
+        private String fieldName;
+        private UserQueryField queryField;
+
+        public Field(String field) {
+            if (field.startsWith(User.SCHEMA)) {
```
tkrille: See the [comment above](https://github.com/osiam/osiam/pull/108/files#r50050541) for groups.

src/main/java/org/osiam/storage/query/UserFilterExpression.java
```diff
@@ -0,0 +1,81 @@
...
+        this.field = new Field(field);
+    }
+
+    @Override
+    public UserFilterExpression.Field getField() {
+        return field;
+    }
+
+    public static class Field implements FilterExpression.Field<UserEntity> {
+        private static final Logger LOG = LoggerFactory.getLogger(UserFilterExpression.Field.class);
```
tkrille: I think [we once agreed](https://github.com/osiam/osiam/issues/40) on how we name loggers.

src/main/java/org/osiam/storage/query/FilterExpression.java
```diff
@@ -49,4 +44,15 @@ public FilterConstraint getConstraint() {
...
         return value;
     }
-}
\ No newline at end of file
+
+    public interface Field<T extends ResourceEntity> {
+
+        String getUrn();
+
+        String getFieldName();
```
tkrille: Why not `getName`?

src/main/java/org/osiam/storage/query/UserFilterExpression.java
```diff
@@ -0,0 +1,81 @@
...
+            } else {
+                fieldName = field.toLowerCase(Locale.ENGLISH);
+            }
+
+            queryField = UserQueryField.fromString(fieldName);
+
+            urn = User.SCHEMA;
+            if (queryField == null) {
+
+                int lastIndexOfPeriod = field.lastIndexOf('.');
```
tkrille, line 48: Ok, I think there's some explanation necessary how and why the algorithm works. Preferably as source comment :)

src/main/java/org/osiam/storage/query/UserFilterExpression.java
tkrille: Because field names are case-insensitive and we are already doing it for core field names some lines above.

src/test/groovy/org/osiam/storage/query/GroupFilterExpressionSpec.groovy
tkrille: I don't think it will issue a warning to the log. The logger call is guarded by `if (queryField == null) {`, which is only the case for non-core fields.

tkrille:
Hey @wallner,

I looked at your recent changes and most stuff is fixed or has been otherwise solved. There are still some issues, but GitHub shows them as "comments on an outdated diff". I replied to some of them, so please take notice of them too.

src/main/java/org/osiam/storage/query/UserSimpleFilterChain.java
wallner, line 53: Yes you are right. I'm not very happy with the nullity though. This is definitely a candidate for optionals in the Java8 PR

src/main/java/org/osiam/storage/query/UserFilterParser.java
wallner: You are right of course. Thank you for taking a look. I fixed it.

src/main/java/org/osiam/storage/query/GroupFilterExpression.java
wallner: I was unable to find a reference to case sensitivity of schemata, my gut feeling would be to be case sensitive but I don't really have a strong opinion on it. What shall we do?

src/main/java/org/osiam/storage/query/UserFilterExpression.java
wallner: How ignorant of me. :-)

src/main/java/org/osiam/storage/query/FilterExpression.java
wallner: You are right. That's a remainder from before the subclass

src/test/groovy/org/osiam/storage/query/GroupFilterExpressionSpec.groovy
wallner: Ok. I reconsidered.

wallner:
Ok. Thank you for reviewing. I think I took care of all your comments

src/main/java/org/osiam/storage/query/GroupFilterExpression.java
tkrille: I'll ask the guys from SCIM what should be the right behavior.

tkrille:
This one's still open: https://github.com/osiam/osiam/pull/108#discussion_r50052470. Also the answer from the SCIM working group is still outstanding. And I need to see some ITs.

src/main/java/org/osiam/storage/query/UserFilterExpression.java
```diff
@@ -0,0 +1,84 @@
...
+    public static class Field implements FilterExpression.Field<UserEntity> {
+        private static final Logger logger = LoggerFactory.getLogger(UserFilterExpression.Field.class);
+        public static final String LOG_MESSAGE = "Period (.) used as field separator in %s. This is not SCIM conform and deprecated!";
+        private String urn;
+        private String name;
+        private UserQueryField queryField;
+
+        public Field(String field) {
+            if (field.startsWith(User.SCHEMA)) {
+                if (field.charAt(User.SCHEMA.length()) == '.'){
```
tkrille: Missing space between `)` and `{`

CHANGELOG.md
```diff
@@ -14,6 +14,13 @@
...
+- Fields of the core schemas for user and group can be fully qualified, ie
```
tkrille: This should be "i.e."

src/main/java/org/osiam/storage/query/UserSimpleFilterChain.java
tkrille, line 53: So, the Java 8 PR has been merged. Do want to migrate this code to `Optional` now?

src/main/java/org/osiam/storage/query/UserFilterExpression.java
```diff
@@ -0,0 +1,93 @@
...
+
+            urn = User.SCHEMA;
+            // we can't recognize a core schema field, it's an extension.
+            if (queryField == null) {
+
+                // lastIndexOf returns -1 if char can not be found in the string
+                int lastIndexOfPeriod = field.lastIndexOf('.');
+                int lastIndexOfColon = field.lastIndexOf(':');
+
+                // the last string part of the string is separated by a colon, that is not SCIM conform but has been
```
tkrille: I think you mean "period".

src/main/java/org/osiam/storage/query/UserFilterExpression.java
```diff
@@ -0,0 +1,93 @@
...
+                // supported by OSIAM for a long time so we generate a warning. As soon as we support extended attributes
+                // for extensions this needs to go.
+                if (lastIndexOfPeriod > lastIndexOfColon) {
+                    urn = field.substring(0, lastIndexOfPeriod);
+                    name = field.substring(lastIndexOfPeriod + 1);
+                    logger.warn(
+                            String.format("Period (.) used as field separator in %s. This is not SCIM conform and deprecated!",
+                                    field));
+                    // a colon is separating the last part of this string and the field of the extension from it's URN.
+                    // Just as it's supposed to be.
```
tkrille, line 62: Don't these comment belong in the `else if` branch?

src/main/java/org/osiam/storage/query/UserFilterExpression.java
```diff
@@ -0,0 +1,93 @@
...
+                } else if (lastIndexOfColon > lastIndexOfPeriod) {
+                    urn = field.substring(0, lastIndexOfColon);
+                    name = field.substring(lastIndexOfColon + 1);
+                } else {
+                    // well, this is awkward.
```
tkrille: I think it's just a user error, i.e. neither "." nor ":" have been found in the fully-qualified attribute name. This is not awkward at all :)

src/main/java/org/osiam/storage/query/UserFilterExpression.java
```diff
@@ -0,0 +1,93 @@
...
+            }
+
+            // Try to resolve a core schema field.
+            queryField = UserQueryField.fromString(name);
+
+            urn = User.SCHEMA;
+            // we can't recognize a core schema field, it's an extension.
+            if (queryField == null) {
+
+                // lastIndexOf returns -1 if char can not be found in the string
```
tkrille, line 47: By documenting the algorithm I rather meant describing the high level thing we are doing here (measuring the length of the last parts separated by "." or ":") and why this works (no ":" and "." are allowed in attribute names), you know?

PR #113, opened 2016-01-21
Raise required Java version to Java 8

### LINKED ISSUE #10: Raise supported minimum java version
Currently [JRE 7 is the minimum version](https://github.com/osiam/osiam/blob/master/docs/detailed-reference-installation.md#requirements) to run OSIAM, but I think we should raise it to JRE 8 to make it possible to use Java 8 features... and [Java 7 is EOL since may 2015](https://java.com/en/download/faq/java_7.xml).

wallner:
This PR raises the minimally required version of Java to 8. It updates the pom and the documentation, including the `CHANGELOG.md` but doesn't change any code yet. I intend to do that in other PRs to keep them managable. This closes #10.

### DIFF
--- a/docs/detailed-reference-installation.md
+++ b/docs/detailed-reference-installation.md
@@ -29,11 +29,11 @@ This document takes the following prerequisites as basis:
     <dt>Operating Systems</dt>
     <dd>Any OS that supports running Java applications</dd>
     <dt>Runtime Environment</dt>
-    <dd>Java 7, OpenJDK 7 JRE</dd>
+    <dd>Java 8, OpenJDK 8 JRE</dd>
     <dt>Application Server</dt>
     <dd>Tomcat 7</dd>
     <dt>Database</dt>
-    <dd>PostgreSQL 9.3 or MySQL 5.7</dd>
+    <dd>PostgreSQL 9.4 or MySQL 5.7</dd>
 </dl>
 
 All of the items have to be installed and running. For the installation of OSIAM
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Changes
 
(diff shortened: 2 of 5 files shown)


docs/detailed-reference-installation.md
```diff
@@ -29,11 +29,11 @@ This document takes the following prerequisites as basis:
     <dt>Operating Systems</dt>
     <dd>Any OS that supports running Java applications</dd>
     <dt>Runtime Environment</dt>
-    <dd>Java 7, OpenJDK 7 JRE</dd>
+    <dd>Java 8, OpenJDK 8 JRE</dd>
     <dt>Application Server</dt>
     <dd>Tomcat 7</dd>
     <dt>Database</dt>
-    <dd>PostgreSQL 9.3 or MySQL 5.7</dd>
+    <dd>PostgreSQL 9.4 or MySQL 5.7</dd>
```
tkrille, line 36: Updating Postgres is not really in the scope of this PR.

wallner, line 36: Well, you are right, but given the upgrade of Postgres in the docker images made it necessary to document it here that we are now testing wih an upgraded version of it. I think it's perfectly ok to keep this one line in this PR and would rather not change that now. After all we are talking about one line of documentation.

tkrille, line 36: It's just that the commit message does not reflect all the changes made. I'm totally fine with just mentioning the postgres update somewhere in the commit. Doesn't need to be a single commit. But I'd be happier with 2 commits.

wallner, line 36: Ok, I made some clarifying additions to the commit message for this PR to make the reason for the documented PostgreSQL version more obvious.

tkrille, line 36: Thanks!

PR #115, opened 2016-01-25
Skip generated classes and migrations for coverage

wallner:
Ignore generated files and database schema migrations when generating
test coverage reports.


tkrille:
What about the generated ANTLR stuff?

wallner:
I forgot about them indeed. Fixed now. Do we need to ignore more?

tkrille:
I think that these are the only generated and otherwise un-testable classes. thanks.

PR #118, opened 2016-01-25
Move default properties to default config file

tkrille:
This way they can be centrally managed using a reasonable configuration
file format (YAML), which seems to be better than using code.

Also add a default config for logging, which wasn't configured before.


wallner:
Thank you

PR #120, opened 2016-01-27
Fix reporting of test coverage

wallner:
The environent variables set in circleci are not available when building pull requests, so this pull request makes sure we don't try to report test coverage data to external systems because we would not be able to get our data there.


dacrome:
Thanks!

PR #123, opened 2016-02-02
Remove unneeded attributes from login template

tkrille:
`th:error` isn't even a valid Thymeleaf attribute and `th:class` is set
to a literal string that is the same as in `class`.


wallner:
Good catch! Thank you!

PR #124, opened 2016-02-03
Refactor LoginController

tkrille:
`OsiamCachingAuthenticationFailureHandler` sets an LDAP-related error
message in the session.
Passing the message to the view was some kind of magical before: No one
knows how and why it worked.
Explicitly read the error message from the session now and add it to the
model.
Use `ModelAndView` in `/login/error` to simplify error handling.

Extract common model attributes as `@ModelAttribute` annotated methods.
This removes the duplication of setting it.

Model attribute `locked` isn't used by the login template, so remove it.

### DIFF
--- a/src/main/java/org/osiam/security/controller/LoginController.java
+++ b/src/main/java/org/osiam/security/controller/LoginController.java
@@ -23,17 +23,23 @@
 
 package org.osiam.security.controller;
 
+import com.google.common.base.Strings;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Controller;
-import org.springframework.ui.Model;
+import org.springframework.web.bind.annotation.ModelAttribute;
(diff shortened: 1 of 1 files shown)


src/main/java/org/osiam/security/controller/LoginController.java
```diff
@@ -23,18 +23,25 @@
 
 package org.osiam.security.controller;
 
+import com.google.common.base.Strings;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Controller;
 import org.springframework.ui.Model;
```
wallner: Codacy Issue found: [Avoid unused imports such as 'org.springframework.ui.Model'](https://www.codacy.com/app/OSIAM/osiam/file/1964679516/issues/source?bid=3080997&fileBranchId=3112117#l30)

tkrille: Fixed

wallner:
Thank you.

PR #125, opened 2016-02-03
Reply with 400 BAD REQUEST to invalid filters

### LINKED ISSUE #66: Search filter cuts string query, when using not supported operator and fired no exception
Moved from https://github.com/osiam/resource-server/issues/32:

> When using = instead of eq in the string filter with the OsiamConnector, the filter string will be cut and no exception fired.
> 
> Right filter string:
> "userName eq \"user1\" and name.formatted eq \"formatted1\""
> -> both constraints will be used
> 
> Wrong filter strings:
> "userName eg \"user1\" and name.formatted = \"formatted1\""
> -> just uses the first constraint, the second will be cut
> 
> "userName = \"user1\" and name.formatted = \"formatted1\""
> -> exception fired: org.osiam.client.exception.ConflictException: no viable alternative at input 'userName"user1"'

You can actually see that ANTLR complains about the wrong operator in the server logs. Seems that the message is put on STDOUT or STDERR.

wallner:
According to the Scim specification invalid filters should receive a
response wit the status code 400 BAD REQUEST and not 409 CONFLICT. This
is fixed by this commit. It replaces all IllegalArgumentExceptions that
were generated by invalid filters with a InvalidFilterExceptions.

Please be aware that this PR requires the corresponding PR for the
integration tests to be accepted as well. Closes #66

### DIFF
--- a/src/main/java/org/osiam/storage/helper/NumberPadder.java
+++ b/src/main/java/org/osiam/storage/helper/NumberPadder.java
@@ -56,7 +56,11 @@ public class NumberPadder {
             throw new IllegalArgumentException("The given value has more than " + (PAD_LENGTH - 1) + " digits.");
         }
 
-        integralPart = new BigInteger(integralPart).add(BIG_OFFSET).toString();
+        try {
+            integralPart = new BigInteger(integralPart).add(BIG_OFFSET).toString();
+        } catch (NumberFormatException e) {
+            throw new IllegalArgumentException("Unable to parse a number from " + value);
+        }
         integralPart = Strings.padStart(integralPart, PAD_LENGTH, '0');
 
         return integralPart + fractionalPart;
--- a/src/main/java/org/osiam/storage/query/ExtensionQueryField.java
+++ b/src/main/java/org/osiam/storage/query/ExtensionQueryField.java
@@ -23,6 +23,7 @@
 
 package org.osiam.storage.query;
 
+import org.osiam.resources.exception.InvalidFilterException;
 import org.osiam.resources.scim.ExtensionFieldType;
 import org.osiam.storage.entities.*;
 import org.osiam.storage.helper.NumberPadder;
@@ -48,7 +49,11 @@ public class ExtensionQueryField {
         if (constraint != FilterConstraint.PRESENT && (field.getType() == ExtensionFieldType.INTEGER ||
                 field.getType() == ExtensionFieldType.DECIMAL)) {
 
-            value = numberPadder.pad(value);
+            try {
+                value = numberPadder.pad(value);
+            } catch (IllegalArgumentException ex) {
+                throw new InvalidFilterException(ex.getMessage());
+            }
         }
 
         final SetJoin<UserEntity, ExtensionFieldValueEntity> join = createOrGetJoin(
--- a/src/main/java/org/osiam/storage/query/OsiamAntlrErrorListener.java
+++ b/src/main/java/org/osiam/storage/query/OsiamAntlrErrorListener.java
@@ -23,10 +23,7 @@
 
 package org.osiam.storage.query;
 
-import java.util.BitSet;
-
 import org.antlr.v4.runtime.ANTLRErrorListener;
-import org.antlr.v4.runtime.InputMismatchException;
 import org.antlr.v4.runtime.Parser;
 import org.antlr.v4.runtime.RecognitionException;
 import org.antlr.v4.runtime.Recognizer;
@@ -34,6 +31,9 @@ import org.antlr.v4.runtime.atn.ATNConfigSet;
 import org.antlr.v4.runtime.dfa.DFA;
 import org.antlr.v4.runtime.misc.NotNull;
 import org.antlr.v4.runtime.misc.Nullable;
+import org.osiam.resources.exception.InvalidFilterException;
+
+import java.util.BitSet;
 
 /**
  * Handles parsing errors.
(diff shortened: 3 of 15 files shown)


src/main/java/org/osiam/storage/helper/NumberPadder.java
```diff
@@ -56,7 +56,11 @@ public String pad(String value) {
             throw new IllegalArgumentException("The given value has more than " + (PAD_LENGTH - 1) + " digits.");
         }
 
-        integralPart = new BigInteger(integralPart).add(BIG_OFFSET).toString();
+        try {
+            integralPart = new BigInteger(integralPart).add(BIG_OFFSET).toString();
+        } catch (NumberFormatException e) {
+            throw new IllegalArgumentException("Unable to parse a number from " + value);
```
tkrille, line 62: Shouldn't we preserve the cause?

src/main/java/org/osiam/storage/query/ExtensionQueryField.java
```diff
@@ -48,7 +49,11 @@ public Predicate addFilter(Root<UserEntity> root, FilterConstraint constraint,
         if (constraint != FilterConstraint.PRESENT && (field.getType() == ExtensionFieldType.INTEGER ||
                 field.getType() == ExtensionFieldType.DECIMAL)) {
 
-            value = numberPadder.pad(value);
+            try {
+                value = numberPadder.pad(value);
+            } catch (IllegalArgumentException ex) {
+                throw new InvalidFilterException(ex.getMessage());
```
tkrille, line 55: Shouldn't we preserve the cause?

src/main/java/org/osiam/storage/query/OsiamAntlrErrorListener.java
```diff
@@ -42,55 +42,52 @@
...
+     * <p>
      * The method is overridden to throw an exception if the parsed input has an invalid syntax.
      */
     @Override
     public void syntaxError(Recognizer<?, ?> recognizer, @Nullable Object offendingSymbol, int line,
-            int charPositionInLine, String msg, @Nullable RecognitionException e) {
+                            int charPositionInLine, String msg, @Nullable RecognitionException e) {
 
-        String errorMessage = msg;
-        if (e instanceof InputMismatchException && msg.endsWith(" expecting VALUE")) {
```
tkrille, line 53: Why did you remove the complete if statement?

src/main/java/org/osiam/storage/query/QueryFilterParser.java
```diff
@@ -38,10 +36,15 @@ public ParseTree getParseTree(String filter) {
         if (Strings.isNullOrEmpty(filter)) {
             return null;
         }
+        OsiamAntlrErrorListener errorListener = new OsiamAntlrErrorListener();
 
         LogicalOperatorRulesLexer lexer = new LogicalOperatorRulesLexer(new ANTLRInputStream(filter));
+        lexer.removeErrorListeners();
+        lexer.addErrorListener(errorListener);
```
tkrille, line 43: This needs some explanation. Has it to do with https://github.com/osiam/osiam/pull/125#discussion_r51790934?

src/main/java/org/osiam/storage/query/QueryFilterParser.java
```diff
@@ -38,10 +36,15 @@ public ParseTree getParseTree(String filter) {
...
+
         LogicalOperatorRulesParser parser = new LogicalOperatorRulesParser(new CommonTokenStream(lexer));
-        parser.addErrorListener(new OsiamAntlrErrorListener());
+        parser.removeErrorListeners();
+        parser.addErrorListener(errorListener);
```
tkrille, line 47: See https://github.com/osiam/osiam/pull/125/files#r51791045

src/main/java/org/osiam/storage/helper/NumberPadder.java
wallner, line 62: As throwable?

tkrille, line 62: I dunno. As cause and using the constructor. Like every other exception in Java.

src/main/java/org/osiam/storage/query/ExtensionQueryField.java
wallner, line 55: I dont' think we are doing anything with the cause further up the stack, so I don't really think keeping the cause wins us anything. Do you see it differently?

tkrille, line 55: It might help to track down bugs when you see the usual giant stacktrace in the log. Does this error case even get logged?

src/main/java/org/osiam/storage/query/OsiamAntlrErrorListener.java
wallner, line 53: Since I'm registering the `OsiamErrorListener` for both the parser and the lexer the error is no longer triggered by the `InputMissmatchException`. We need to register an error handler for both, the lexer and the parser, otherwise we could not get hold of the lexer error.

src/main/java/org/osiam/storage/helper/NumberPadder.java
tkrille, line 62: Never mind, the error won't be logged anyway.

src/main/java/org/osiam/storage/query/ExtensionQueryField.java
tkrille, line 55: Never mind, the error won't be logged anyway.

src/main/java/org/osiam/storage/query/OsiamAntlrErrorListener.java
tkrille, line 53: Does this correlate with https://github.com/osiam/osiam/pull/125/files#diff-6e7b8e789137b2ead84ff72590e89965R79?

src/main/java/org/osiam/storage/helper/NumberPadder.java
wallner, line 62: We could discuss logging this error as debug or something. I think this would be a different pull request.

PR #127, opened 2016-02-03
Load configuration and assets from the file system

tkrille:
Add notion of `OSIAM_HOME`, an arbitrary folder in the file system that
contains the static assets and the configuration.

The home directory will default to `$HOME/.osiam`, if not specified.
This seems to be a good place, as the current user will certainly have
access to this folder.
It's also a "best practice" by many other applications.

Use a `ReloadableResourceBundleMessageSource`, because it's the only one
that can load message source bundles from the file system.


tkrille:
You can start OSIAM on the command line like this:

```
mvn spring-boot:run -Drun.arguments='--osiam.home=src/main/deploy'
OSIAM_HOME=src/main/deploy java -jar target/osiam.war
java -jar target/osiam.war --osiam.home=src/main/deploy
java -Dosiam.home=src/main/deploy -jar target/osiam.war
```

tkrille:
ITs: https://github.com/osiam/connector4java-integration-tests/pull/321
Docker image: https://github.com/osiam/docker-image/pull/33

wallner:
That's very good work. Thank you!

PR #130, opened 2016-02-10
Extract assets and configuration on startup

tkrille:
Assets and configuration will copied from the classpath to the home
directory, when starting OSIAM.

The contents of the home directory have been put on the classpath,
because it's the only place that is working under the different runtime
environments, e.g. Maven (with Spring Boot plugin), Tomcat (or other
Servlet containers) and standalone usage with `java -jar osiam.war`.

Initialization will not be done, if the home directory already contains
files or folders, to prevent accidentally overwriting a customized
setup.

Also load the default configuration that's contained in the classpath,
to work around a chicken-and-egg problem:
If OSIAM starts without a configuration in place, it will not reach the
point, where it initializes the home directory, because property
replacements will fail beforehand.
Always loading the default config, which is the same config that will be
copied to the home directory anyway, makes sure all properties are set
to working defaults.
On the next start the values from the config file in the home directory
will override the ones defined in the default config.

Remove the distribution creation and make the `.war` file the single
artifact to distribute.
The `.war.` file contains all needed assets now, so a separate
distribution artifact is not needed anymore.
Also add the changelog and readme to the `.war.` file from now on, as
they have been in included in the distribution artifact before.
Assets and default configuration can still be extracted from the `.war.`
file, as it's just a ZIP file, but from a different path.


dacrome:
Works like a charm! Thank you very much! This is a huge step to OSIAM 3.0!!! :dancers:

PR #132, opened 2016-02-11
Remove ability to search by a user's password

### LINKED ISSUE #56: Remove ability to search by a user's password
This was needed so that the split auth-server can verify the user's credentials. Not needed anymore after the merge. It's also not specified by SCIM and afaic a potential security risk. There's also some logic that determines whether the search did not produce any result, leading to a `Thread.sleep` of 1s, which must be removed too.

tkrille:
This was needed so that the split auth-server can verify the user's
credentials.
Not needed anymore after the merge.

Resolves #56


tkrille:
Changed ITs: osiam/connector4java-integration-tests#322

dacrome:
Thanks!

Issue #48, opened 2015-10-23
Check exception handling


tkrille:
The merge caused the exception handling to change. Instead of 1 exception handler for all controllers, we now have 2 for specific controllers. One exception handler processes faults from the REST resources and always returns JSON. The other processes all faults from the web resources, i.e. HTML pages, and returns the default OAuth error view. It also only contains a handler for `Exception.class`.

The problems with that are:
- Faults from `@FrameworkEndpoint`s will not be handled anymore by our exception handler, but will trigger the whitelabel error page included with Spring Boot
- Both handlers are responsible for handling `Exception.class`, which already led to issues on different machines (https://github.com/osiam/osiam/pull/44#issuecomment-150341982)
- The distinction between web and REST resources is not fully implemented. For example, some RESTful resources like token validation will be handled by the web exception handler, though they always respond with JSON.

So, we have to find a solution for this.

tkrille:
> Faults from @FrameworkEndpoints will not be handled anymore by our exception handler, but will trigger the whitelabel error page included with Spring Boot

I've checked this point and, fortunately, we already implement [`/oauth/error`](https://github.com/osiam/osiam/blob/master/src/main/java/org/osiam/security/controller/AccessConfirmationController.java#L71). So, no worries here. I've also actually tested that errors from `/oauth/authorize` lead to the well-know error screen:

![oauth_error](https://cloud.githubusercontent.com/assets/5213491/15835419/2d9aa44a-2c30-11e6-99b8-ce667bfb60d5.png)

Maybe we change this in the future ... :)

tkrille:
> Both handlers are responsible for handling Exception.class, which already led to issues on different machines

This should be no problem anymore, because responsibility of the exception handler has been restricted.

dacrome:
So what is the concrete doing here? Seperation of web and rest controller exception handler?

tkrille:
> The distinction between web and REST resources is not fully implemented. For example, some RESTful resources like token validation will be handled by the web exception handler, though they always respond with JSON.

See #225

tkrille:
@dacrome I've already done it :)

dacrome:
;)

tkrille:
You wanna review #225? :)

PR #134, opened 2016-02-11
Replace homemade validation with JSR303 validation

### LINKED ISSUE #62: Resources with only 'primary' sub-attribute set should be rejected
Moved from https://github.com/osiam/resource-server/issues/10:

Currently it is possible to create a user with multi-valued attributes where only the sub-attribute 'primary' is set. This should not be possible, a multi-valued attribute should only be valid if at least one of the sub-attributes are set that needed for the identification of the resource.

wallner:
This commit replaces the homemade validation with Springs JSR303 bean
valdation. It also replaces all direkt manipulations of
HttpServletRequest and HttpServletResponse objects with the saner Spring
alternatives. It also introduces MocMVC tests. Closes #62


src/main/java/org/osiam/resources/scim/MultiValuedAttribute.java
```diff
@@ -28,6 +28,8 @@
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.base.Strings;
+import org.hibernate.validator.constraints.NotBlank;
+import org.hibernate.validator.constraints.NotEmpty;
```
wallner: Codacy Issue found: [Avoid unused imports such as 'org.hibernate.validator.constraints.NotEmpty'](https://www.codacy.com/app/OSIAM/osiam/file/2030369025/issues/source?bid=3080997&fileBranchId=3128707#l32)

src/main/java/org/osiam/resources/scim/Email.java
```diff
@@ -24,8 +24,10 @@
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import org.hibernate.validator.constraints.NotBlank;
 import org.osiam.resources.exception.SCIMDataValidationException;
 
+import javax.validation.constraints.NotNull;
```
wallner: Codacy Issue found: [Avoid unused imports such as 'javax.validation.constraints.NotNull'](https://www.codacy.com/app/OSIAM/osiam/file/2036403903/issues/source?bid=3080997&fileBranchId=3128707#l30)

tkrille: Do you want to remove the unused import?

pom.xml
```diff
@@ -245,11 +245,24 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-test</artifactId>
+            <version>4.2.1.RELEASE</version>
```
dacrome: You can remove the version tag because spring boot brings the dependency and the current version `4.2.4.RELEASE`

src/main/java/org/osiam/configuration/JacksonConfig.java
```diff
@@ -0,0 +1,20 @@
...
+import com.fasterxml.jackson.databind.Module;
+import com.fasterxml.jackson.databind.module.SimpleModule;
+import org.osiam.resources.helper.UserDeserializer;
+import org.osiam.resources.scim.User;
+import org.springframework.boot.autoconfigure.web.WebMvcAutoConfiguration;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class JacksonConfig extends WebMvcAutoConfiguration {
```
dacrome: Please add empty line after class declaration

src/main/java/org/osiam/configuration/JacksonConfig.java
```diff
@@ -0,0 +1,20 @@
...
+    @Bean
+    public Module jacksonBuilder() {
+        return new SimpleModule("userDeserializerModule", new Version(1, 0, 0, null, "org.osiam",
+                "scim-schema"))
```
dacrome: I think this line can be append to the line before

src/main/java/org/osiam/configuration/JacksonConfig.java
tkrille: Don't inherit from `WebMvcAutoConfiguration`. I don't think it's a good idea. Why is it necessary anyway? The docs state:

> Another way to customize Jackson is to add beans of type com.fasterxml.jackson.databind.Module to your context. They will be registered with every bean of type ObjectMapper, providing a global mechanism for contributing custom modules when you add new features to your application.

(http://docs.spring.io/spring-boot/docs/1.3.2.RELEASE/reference/htmlsingle/#howto-customize-the-jackson-objectmapper)

So it should be enough to add the `@Bean` declaration to `OAuth2ResourceServerConfig` or some other place (or just leave it in this `@Configuration` class).

src/main/java/org/osiam/resources/controller/ServiceProviderConfigsController.java
```diff
@@ -50,7 +48,7 @@ public ServiceProviderConfig getConfig() {
         public static final String SCHEMA = "urn:ietf:params:scim:schemas:core:2.0:ServiceProviderConfig";
         public final Supported patch = new Supported(true);
         public final Supported bulk = new BulkSupported(false);
-        public final Supported filter = new FilterSupported(true, RequestParamHelper.MAX_RESULT);
+        public final Supported filter = new FilterSupported(true, 100);
```
dacrome: Please do not use magic number, create `MAX_RESULT` in this class

src/main/java/org/osiam/resources/scim/Email.java
```diff
@@ -68,6 +69,8 @@ public String getOperation() {
     }
 
     @Override
+    @NotBlank(message = "Mulit-Valued attributes may not have empty values")
```
timopick: Typo "Mulit" / 9 more in the commit

src/main/java/org/osiam/configuration/JacksonConfig.java
```diff
@@ -0,0 +1,20 @@
...
+import org.osiam.resources.helper.UserDeserializer;
+import org.osiam.resources.scim.User;
+import org.springframework.boot.autoconfigure.web.WebMvcAutoConfiguration;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class JacksonConfig extends WebMvcAutoConfiguration {
+    @Bean
+    public Module jacksonBuilder() {
```
tkrille: Please name the method somehow after the bean you creating, e.g. `userDeserializerModule` or `jacksonUserDeserializerModule`. In the end, the method name becomes the bean id.

pom.xml
```diff
@@ -245,11 +245,24 @@
...
         <dependency>
             <groupId>com.jayway.restassured</groupId>
             <artifactId>json-path</artifactId>
-            <version>2.5.0</version>
+            <version>2.8.0</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.jayway.jsonpath</groupId>
+            <artifactId>json-path</artifactId>
```
tkrille, line 264: Why do we need the other json-path, too? We should agree on one.

src/main/java/org/osiam/resources/controller/ServiceProviderConfigsController.java
tkrille: Should be configurable anyways, but that's out of scope. AFAIK we do not even enforce this limit on the service level. 

@dacrome: where should this constant belong? `RequestParamHelper` does not exist anymore.

src/main/java/org/osiam/resources/controller/GroupController.java
```diff
@@ -95,46 +72,41 @@ public void delete(@PathVariable final String id) {
...
     @RequestMapping(method = RequestMethod.GET)
     @ResponseBody
-    public SCIMSearchResult<Group> searchWithGet(HttpServletRequest request) {
-        Map<String, Object> parameterMap = requestParamHelper.getRequestParameterValues(request);
-        SCIMSearchResult<Group> scimSearchResult = scimGroupProvisioning.search((String) parameterMap.get("filter"),
-                (String) parameterMap.get("sortBy"), (String) parameterMap.get("sortOrder"),
-                (int) parameterMap.get("count"), (int) parameterMap.get("startIndex"));
-
-        return attributesRemovalHelper.removeSpecifiedAttributes(scimSearchResult, parameterMap);
+    public SCIMSearchResult<Group> searchWithGet(@RequestParam Map<String, String> parameterMap) {
```
tkrille: I don't like `parameterMap` as name. Something like `requestParams` seems better to me. You don't have to encode the type in the name.

tkrille: This is also true for the other controllers.

dacrome:
You can delete `JsonInputValidator` and `JsonInputValidatorSpec` because it's not used anymore.

tkrille:
The ITs are completely failing with messages like this:

```
Stacktrace was: org.osiam.client.exception.ConnectionInitializationException: Could not deserialize the error response for the HTTP status 'Method Not Allowed'. Original response: {"timestamp":1455529857964,"status":405,"error":"Method Not Allowed","exception":"org.springframework.web.HttpRequestMethodNotSupportedException","message":"Request method 'POST' not supported","path":"/osiam/oauth/token"}
```

This seems very odd to me. Why in the world is `POST` not allowed for `/osiam/oauth/token`?

src/main/java/org/osiam/configuration/JacksonConfig.java
wallner: The `@Bean` doesn't get picked up if the `@Configuration` doesn't inherit from `WebMvcAutoConfiguration` that's the reason I put the Jackson configuration in it's own `@Configuration` class in the first place.

pom.xml
wallner, line 264: I added jsonpath because the MockMVC makes good use of it.

src/main/java/org/osiam/resources/controller/GroupController.java
wallner: I agree with you.

src/main/java/org/osiam/resources/controller/ServiceProviderConfigsController.java
dacrome: @tkrille Like I wrote: `...create MAX_RESULT in this class`

src/main/java/org/osiam/configuration/JacksonConfig.java
tkrille: Unfortunately, I cannot test this assumptions. All I know is from the official docs and they don't say anything about such a restriction. Inheriting the auto configuration class just seems unnecessary.

pom.xml
tkrille, line 264: So MockMVC depends on this json-path lib?

src/main/java/org/osiam/resources/controller/ServiceProviderConfigsController.java
tkrille: Oh, I see. Sounds viable.

src/main/java/org/osiam/configuration/JacksonConfig.java
wallner: I tried. If the `@Configuration` is not a sub class of `WebMvcAutoConfiguration` the `@Bean` of the type `Module` doesn't get picked up.

pom.xml
wallner, line 264: Only if I want to test json expressions like this:

```
response.andExpect(status().isOk())
    .andExpect(jsonPath('$.id').value(uuid as String))
```

I think that's a very good feature to have available in our test harness.

src/main/java/org/osiam/resources/controller/ServiceProviderConfigsController.java
wallner: I think the constant does belong in the `ScimSearchResult` and already spoke with @dacrome about it. I will provide a follow-up PR after this one is merged.

src/main/java/org/osiam/configuration/JacksonConfig.java
tkrille: Ok, as soon as the ITs are running, I will try to confirm this.

tkrille: Ok, I've debugged the startup and `com.fasterxml.jackson.databind.ObjectMapper#registerModule` will be called with our module as parameter, even if the inheritance from `WebMvcAutoConfiguration` is gone.

So I think, it can be safely removed.

wallner: I agree, I removed it.

wallner:
I'm not seeing the ITs fail. I'm sorry.

pom.xml
```diff
@@ -245,11 +245,23 @@
...
+        </dependency>
+
         <!-- Testing JSON -->
         <dependency>
             <groupId>com.jayway.restassured</groupId>
             <artifactId>json-path</artifactId>
-            <version>2.5.0</version>
+            <version>2.8.0</version>
+            <scope>test</scope>
+        </dependency>
```
dacrome, line 260: please add empty line between the dependency tags

pom.xml
tkrille, line 264: But this is already provided by the other json-path, that seems to be a test dependency for some time. I cannot remember when we added it or how and where we even use it :)

tkrille, line 264: Oh, I see. The other json-path has been introduced when you merged the connector / scim-schema into the server. It had been used for the de-/serialization tests in the scim-schema. So I think one is enough. Can rewrite the tests to use the one provided by rest-assured. At least, this is lib for tests. I think the other one is more suited for production.

wallner, line 264: The other json-path came in with the merge of the scim-schema and is used there in tests. I'm happy to remove it in a different PR, the old one that is.

tkrille, line 264: Alright, I get it now :) never mind.

src/main/java/org/osiam/configuration/JacksonConfig.java
```diff
@@ -0,0 +1,20 @@
...
+import org.springframework.boot.autoconfigure.web.WebMvcAutoConfiguration;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class JacksonConfig {
+
+    @Bean
+    public Module jacksonUserDeserializerModule() {
+        return new SimpleModule("userDeserializerModule", new Version(1, 0, 0, null, "org.osiam", "scim-schema"))
```
tkrille: `scim-schema`? I think ist must be `osiam` now.

wallner: Indeed. Should have been for quite some time.

src/main/java/org/osiam/resources/controller/GroupController.java
```diff
@@ -95,46 +72,41 @@ public void delete(@PathVariable final String id) {
...
-    public SCIMSearchResult<Group> searchWithPost(HttpServletRequest request) {
-        Map<String, Object> parameterMap = requestParamHelper.getRequestParameterValues(request);
-        SCIMSearchResult<Group> scimSearchResult = scimGroupProvisioning.search((String) parameterMap.get("filter"),
-                (String) parameterMap.get("sortBy"), (String) parameterMap.get("sortOrder"),
-                (int) parameterMap.get("count"), (int) parameterMap.get("startIndex"));
+    public SCIMSearchResult<Group> searchWithPost(@RequestParam Map<String, String> requestParameters) {
+        SCIMSearchResult<Group> scimSearchResult = scimGroupProvisioning.search(requestParameters.get("filter"),
+                requestParameters.get("sortBy"),
+                requestParameters.getOrDefault("sortOrder", "ascending"),
+                Integer.parseInt(requestParameters.getOrDefault("count", "100")),
```
tkrille: What about the magic number (`100`) now? Do we want to add it as a constant to `SCIMSearchResult` or some other class? As I said, it should be configurable anyway, so I don't really care. /cc @dacrome

src/main/java/org/osiam/resources/controller/ServiceProviderConfigController.java
```diff
@@ -23,20 +23,20 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import org.springframework.stereotype.Controller;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.ResponseBody;
 
-import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import java.util.HashSet;
+import java.util.Set;
 
 @Controller
```
tkrille: Do want to make this a `@RestController`, too, and remove the `@ResponseBody` below?

src/main/java/org/osiam/resources/controller/GroupController.java
```diff
@@ -95,46 +72,41 @@ public void delete(@PathVariable final String id) {
...
-        Group group = jsonInputValidator.validateJsonGroup(request);
-        Group createdGroup = scimGroupProvisioning.update(id, group);
-        setLocation(request, response);
-        return createdGroup;
+        Group updatedGroup = scimGroupProvisioning.update(id, group);
+        return buildResponseWithLocation(updatedGroup, builder, HttpStatus.OK);
     }
 
     @RequestMapping(method = RequestMethod.GET)
     @ResponseBody
```
tkrille: This annotation can be removed, it's a `@RestController`: https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/web/bind/annotation/RestController.html

src/main/java/org/osiam/resources/controller/GroupController.java
```diff
@@ -95,46 +72,41 @@ public void delete(@PathVariable final String id) {
...
-                (String) parameterMap.get("sortBy"), (String) parameterMap.get("sortOrder"),
-                (int) parameterMap.get("count"), (int) parameterMap.get("startIndex"));
-
-        return attributesRemovalHelper.removeSpecifiedAttributes(scimSearchResult, parameterMap);
+    public SCIMSearchResult<Group> searchWithGet(@RequestParam Map<String, String> requestParameters) {
+        return searchWithPost(requestParameters);
     }
 
     @RequestMapping(value = "/.search", method = RequestMethod.POST)
     @ResponseBody
```
tkrille: See https://github.com/osiam/osiam/pull/134/files#r53046544

src/main/java/org/osiam/resources/controller/UserController.java
```diff
@@ -25,35 +25,31 @@
...
- * <p/>
+ * <p>
  * it is based on the SCIM 2.0 API Specification:
- * <p/>
+ * <p>
  * http://tools.ietf.org/html/draft-ietf-scim-api-00#section-3
- *
- *
  */
 @Controller
```
tkrille: Do want to make this a `@RestController`, too, and remove all the `@ResponseBody`s below?

src/main/java/org/osiam/resources/controller/GroupController.java
dacrome: It's ok, let it like it is.

src/main/java/org/osiam/resources/exception/RestExceptionHandler.java
```diff
@@ -32,18 +32,19 @@
...
 import org.springframework.web.bind.annotation.ResponseBody;
 import org.springframework.web.bind.annotation.ResponseStatus;
-import org.springframework.web.servlet.mvc.method.annotation.ResponseEntityExceptionHandler;
 
 @ControllerAdvice({
         "org.osiam.auth.oauth_client",
         "org.osiam.metrics.controller",
         "org.osiam.resources.controller"
 })
-public class RestExceptionHandler extends ResponseEntityExceptionHandler {
```
tkrille, line 46: Why don't we want to use this as a base class anymore. Seems quite helpful to me:

> A convenient base class for `@ControllerAdvice` classes that wish to provide centralized exception handling across all `@RequestMapping` methods through `@ExceptionHandler` methods.
> This base class provides an `@ExceptionHandler` method for handling standard Spring MVC exceptions that returns a `ResponseEntity` to be written with message converters.

tkrille:
The right SCIM schema won't be enforced anymore. I was able to do the following request and get a successful response:

```
2016-02-16 18:44:59 org.osiam.bundled.org.apache.http.wire [DEBUG] http-outgoing-0 >> "POST /osiam/Users HTTP/1.1[\r][\n]"
...
2016-02-16 18:44:59 org.osiam.bundled.org.apache.http.wire [DEBUG] http-outgoing-0 >> "[\r][\n]"
2016-02-16 18:44:59 org.osiam.bundled.org.apache.http.wire [DEBUG] http-outgoing-0 >> "{"schemas":["urn:scim:schemas:core:2.0:User"],"userName":"newUser"}"
2016-02-16 18:44:59 org.osiam.bundled.org.apache.http.wire [DEBUG] http-outgoing-0 << "HTTP/1.1 201 Created[\r][\n]"
...
2016-02-16 18:44:59 org.osiam.bundled.org.apache.http.wire [DEBUG] http-outgoing-0 << "{"id":"bfdc3ac3-97ed-490e-b7e5-6217f0964b25","meta":{"created":"2016-02-16T18:44:59.850+01:00","lastModified":"2016-02-16T18:44:59.850+01:00","resourceType":"User"},"schemas":["urn:ietf:params:scim:schemas:core:2.0:User"],"userName":"newUser","active":false}[\r][\n]"
...
```

As you can see, I've sent the wrong (legacy) schema `urn:scim:schemas:core:2.0:User`. The server not only accepted this, but also changed it in its response to `urn:ietf:params:scim:schemas:core:2.0:User`, which is the right schema.

tkrille:
Commit Message:

Typo: "direkt" -> "direct"
Typo: "MocMVC" -> "MockMVC"

src/main/java/org/osiam/resources/controller/GroupController.java
wallner: No. Let's move it to ScimSearchResult

src/main/java/org/osiam/resources/controller/ServiceProviderConfigController.java
wallner: I planned doing a second commit for those and add a few tests with it.

src/main/java/org/osiam/resources/controller/UserController.java
wallner: Yes.

src/main/java/org/osiam/resources/controller/ServiceProviderConfigController.java
tkrille: You did it for `GroupController`, too, and it will definitively work:

> `public @interface RestController`
> A convenience annotation that is itself annotated with `@Controller` and `@ResponseBody`.
> Types that carry this annotation are treated as controllers where `@RequestMapping` methods assume `@ResponseBody` semantics by default.

It's just framework magic.

src/main/java/org/osiam/resources/exception/RestExceptionHandler.java
wallner, line 46: I had to remove the super class because Spring was no longer able to unambiguously resolve the exception handler to apply.

tkrille, line 46: Oh, I see. `WebMvcAutoConfiguration` defines one by itself. This is due to our necessary removal of `@EnableWebMVC`.

tkrille, line 46: BTW: The conflicting method is `org.osiam.resources.exception.RestExceptionHandler#handleValidationException`, but I think it will be fine anyways. There's still #48 open, that deals with the exception handling.

src/main/java/org/osiam/resources/exception/RestExceptionHandler.java
```diff
@@ -71,8 +81,7 @@ public ErrorResponse handleResourceExists(ResourceExistsException e) {
 
     @ExceptionHandler(OsiamBackendFailureException.class)
     @ResponseStatus(HttpStatus.INTERNAL_SERVER_ERROR)
-    @ResponseBody
```
tkrille, line 74: I think this one is actually still needed.

src/main/java/org/osiam/resources/controller/ServiceProviderConfigController.java
```diff
@@ -23,22 +23,20 @@
...
 
-import java.util.HashSet;
-import java.util.Set;
-
-import org.osiam.resources.helper.RequestParamHelper;
-import org.springframework.stereotype.Controller;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import org.osiam.resources.scim.SCIMSearchResult;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.ResponseBody;
```
tkrille: Unused import

src/main/java/org/osiam/Osiam.java
```diff
@@ -28,7 +28,6 @@
 import java.util.Map;
 
 @SpringBootApplication
-@EnableWebMvc
```
tkrille, line 31: Please, also remove the import statement.

src/main/java/org/osiam/resources/exception/RestExceptionHandler.java
wallner, line 74: Right. fixed.

src/main/java/org/osiam/resources/controller/MeController.java
```diff
@@ -86,7 +86,6 @@
      * @return an object to represent the json format.
      */
     @RequestMapping(value = "/**", method = { RequestMethod.GET, RequestMethod.POST })
-    @ResponseBody
```
tkrille, line 89: Please, also remove the import.

tkrille:
[Validation of `schemas` attribute](https://github.com/osiam/osiam/pull/134#issuecomment-184797674) will be addressed in an upcoming issue.

tkrille:
Thanks a lot!

dacrome:
Awesome work and discussion guys!!!

PR #135, opened 2016-02-12
Squash Flyway migrations

tkrille:
This unifies the first 3 database versions, that reflect the way the
database schema looks like in the latest 2.x version.
The idea is that version 1 will be the new baseline, so that on an
upgrade from 2.x to 3.0, Flyway will mark this schema as applied
(baselined) and won't run it.
After that, version 2 will be applied, that is currently just changing
the scopes of the well-known clients.
Version 2 should be used for all upcoming automatic changes, that deal
with old data, not updated schemas, like removing the auth-server
client, renaming the auth-server extension, etc.

The PostgreSQL scripts have been re-generated by Hibernate's auto DDL
feature and exported with IntelliJ.
It had one unique constraint missing: `osiam_client.id`.
The MySQL scripts are the same as before, but combined.


dacrome:
Please rebase.

dacrome:
I'm not sure If I get the idea of this PR. What did you tripped over that you made this change? Is it save to upgrade from 2 to 3 (I guess it is)?

tkrille:
rebased

tkrille:
@dacrome: We want to remove the example data from the Flyway scripts. This means we have to amend the first migration version. Because one has to manually migrate the database anyway, it's not good to change something, that has a fixed version, and to just have a fresh start to remedy old mistakes, it just makes no sense to use the old scripts.

dacrome:
Ok thanks for the explanation!

PR #143, opened 2016-02-17
Make sure core schema is present and valid

### LINKED ISSUE #142: Validate schema of incoming resources
The merge of pr #134 disabled the validation of the schema resources. To be SCIM compliant we need to validate the schemas again.

wallner:
The last refactoring of the validation removed the ability to validate
the schema of scim users and groups, i.e. It was possible to provide
users and groups with a wrong or no schema present.

This commit fixes it by providing a custom validator and annotation to
validate the provided schemas.

This closes #142

### DIFF
--- a/src/main/java/org/osiam/resources/scim/Resource.java
+++ b/src/main/java/org/osiam/resources/scim/Resource.java
@@ -24,6 +24,7 @@
 package org.osiam.resources.scim;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.collect.ImmutableSet;
 import org.hibernate.validator.constraints.NotEmpty;
 import org.osiam.resources.exception.SCIMDataValidationException;
 
@@ -103,7 +104,7 @@ public abstract class Resource implements Serializable {
      */
     @NotEmpty
     public Set<String> getSchemas() {
-        return schemas;
+        return ImmutableSet.copyOf(schemas);
     }
 
     @Override
--- a/src/main/java/org/osiam/resources/validation/ValidCoreSchema.java
+++ b/src/main/java/org/osiam/resources/validation/ValidCoreSchema.java
@@ -1,0 +1,43 @@
+/**
+ * The MIT License (MIT)
+ *
+ * Copyright (C) 2013-2016 tarent solutions GmbH
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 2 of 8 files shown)


tkrille:
Why did you close this one?

wallner:
Äh

tkrille:
Haha :D

src/main/java/org/osiam/resources/validation/ValidCoreSchema.java
```diff
@@ -0,0 +1,26 @@
...
+    Class<?>[] groups() default {};
+
+    Class<? extends Payload>[] payload() default {};
+
+    String value();
+
+    @Target({ElementType.METHOD, ElementType.FIELD, ElementType.ANNOTATION_TYPE})
+    @Retention(RetentionPolicy.RUNTIME)
+    @Documented
+    @interface List {
```
tkrille: I don't think that this is needed. It's not used anywhere.

wallner: That's right, it's not needed anywhere, but good practice when writing validators, because it allows the application of lists in the validator. That said: I wouldn't be unhappy when I removed it. Do you think it has to go?

tkrille: Sorry, but I don't understand this. Can you give me an example of an application of `@List`?

tkrille:
I really don't like, that you made [`Resource#schemas` protected](https://github.com/osiam/osiam/pull/143/files#diff-150eaf5df40108479bf067bdefcb99c3R46) and [`Resource#getSchemas` abstract](https://github.com/osiam/osiam/pull/143/files#diff-150eaf5df40108479bf067bdefcb99c3R105). I'd prefer that you override `Resource#getSchemas`, but delegate to `super.getSchemas`.

And doesn't get the validation annotations inherited? Then you could leave the `@NotEmpty` on `Resource#schemas`.

wallner:
Validations get inherited, I'm not sure about your dislike of the abstract method. I need to think about it.

tkrille:
So, did you think about it? :)

src/main/java/org/osiam/resources/validation/ValidCoreSchema.java
tkrille: Do you mean something like [`@Repeatable`](https://docs.oracle.com/javase/tutorial/java/annotations/repeating.html)?

wallner:
I did and I picked up your idea.

tkrille:
Please rebase.

src/main/java/org/osiam/resources/scim/Resource.java
```diff
@@ -103,7 +104,7 @@ public Meta getMeta() {
      */
     @NotEmpty
     public Set<String> getSchemas() {
-        return schemas;
+        return ImmutableSet.copyOf(schemas);
```
tkrille, line 107: It's back again?

wallner, line 107: Yes.

tkrille:
The new classes are missing the correct license header now. Please rebase and fix the header.

wallner:
Yes. Updated.

src/test/groovy/org/osiam/resources/controller/UserControllerSpec.groovy
```diff
@@ -109,6 +109,18 @@ class UserControllerSpec extends Specification {
                 .andExpect(jsonPath('$.status').value('400'))
     }
 
+    def 'Creating an User with invalid Schema raises a 400 BAD REQUEST'() {
```
tkrille: Typo: "a User"

PR #145, opened 2016-02-20
Switch to constructor injection

### LINKED ISSUE #144: Switch to constructor inject instead of field injection

dacrome:
Resolves #144

### DIFF
--- a/src/main/java/org/osiam/OsiamHome.java
+++ b/src/main/java/org/osiam/OsiamHome.java
@@ -10,7 +10,11 @@ import org.springframework.core.io.support.PathMatchingResourcePatternResolver;
 import org.springframework.stereotype.Component;
 
 import java.io.IOException;
-import java.nio.file.*;
+import java.nio.file.DirectoryStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
 
 import static com.google.common.base.Preconditions.checkState;
 
--- a/src/main/java/org/osiam/auth/login/internal/InternalAuthenticationProvider.java
+++ b/src/main/java/org/osiam/auth/login/internal/InternalAuthenticationProvider.java
@@ -44,27 +44,36 @@ import org.springframework.security.core.GrantedAuthority;
 import org.springframework.security.core.authority.SimpleGrantedAuthority;
 import org.springframework.stereotype.Component;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
 @Component
 public class InternalAuthenticationProvider implements AuthenticationProvider,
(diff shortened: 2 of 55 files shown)


src/main/java/org/osiam/Osiam.java
```diff
@@ -2,20 +2,16 @@
 
 import com.google.common.collect.ImmutableMap;
 import com.ryantenney.metrics.spring.config.annotation.EnableMetrics;
-import com.zaxxer.hikari.HikariConfig;
-import com.zaxxer.hikari.HikariDataSource;
+import com.zaxxer.hikari.*;
```
wallner: I'm not fundamentally opposed to wildcard  imports, even though I'm not a big fan of them. But please don't commit files with this PR that don't have any changes other than import collapses.

src/main/java/org/osiam/Osiam.java
```diff
@@ -2,20 +2,16 @@
...
 import org.flywaydb.core.Flyway;
 import org.flywaydb.core.api.MigrationVersion;
-import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.beans.factory.annotation.Value;
+import org.springframework.beans.factory.annotation.*;
```
wallner: Please see above.

src/main/java/org/osiam/Osiam.java
```diff
@@ -2,20 +2,16 @@
...
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.boot.builder.SpringApplicationBuilder;
 import org.springframework.boot.context.web.SpringBootServletInitializer;
 import org.springframework.context.MessageSource;
-import org.springframework.context.annotation.Bean;
-import org.springframework.context.annotation.EnableAspectJAutoProxy;
-import org.springframework.context.annotation.Primary;
+import org.springframework.context.annotation.*;
```
wallner: Please see comment above.

src/main/java/org/osiam/OsiamHome.java
```diff
@@ -1,10 +1,8 @@
 package org.osiam;
 
 import com.google.common.base.Strings;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.beans.factory.annotation.Value;
+import org.slf4j.*;
```
wallner: My comment about wildcard imports from the file `Osiam.java` holds true here as well.

wallner:
I'm happy to merge this, but please take the two files out that only collapse imports.

src/main/java/org/osiam/Osiam.java
dacrome: I don't know why I changed the IntelliJ settings yesterday to use wildcard imports, but I'm also not a fan. I would like to introduce this as a default coding style for all osiam repositories.

dacrome: So I meant to introduce single class import as [coding convention](https://github.com/osiam/osiam/blob/master/CONTRIBUTING.md#coding-conventions).

wallner:
Thank you!

PR #146, opened 2016-02-20
Refactor configurations

tkrille:
This moves all configurations from `org.osiam.Osiam` to their dedicated
configuration classes.
See the individual commits for further details.


wallner:
Thank You.

PR #147, opened 2016-02-21
Use @JsonFilter instead of Mix-In

dacrome:
To make sure Spring's object mapper is also usable
in the `AttributesRemovalHelper`, this PR get rid off
Jackson's Mix-In and introduce `@JsonFilter`.
This is only possible  by configuring on startup a
filter which does no property filtering.


dacrome:
Hey @tkrille,
please run this change with the ITs, because you are the only one whose IT test order is showing that the current master is broken :)

src/main/java/org/osiam/resources/helper/AttributesRemovalHelper.java
```diff
@@ -43,6 +55,11 @@
 
     private ObjectMapper mapper = new ObjectMapper();
```
tkrille, line 56: Remove the constructor call and make this field `private`.

src/main/java/org/osiam/resources/helper/AttributesRemovalHelper.java
```diff
@@ -135,22 +152,21 @@
     }
 
     private ObjectWriter getObjectWriter(List<String> fieldsToReturn) {
+        if (fieldsToReturn.isEmpty()) {
+            FilterProvider filter = new SimpleFilterProvider().addFilter("filter properties by name", SimpleBeanPropertyFilter.serializeAll());
```
tkrille: Do we really need another no-op filter? Shouldn't the one defined in the global `ObjectMapper` be enough?

src/main/java/org/osiam/resources/helper/AttributesRemovalHelper.java
dacrome, line 56: Thanks, good catch!

src/main/java/org/osiam/resources/helper/AttributesRemovalHelper.java
dacrome: Yes, good catch again

dacrome:
I close this PR because @wallner created a [better solution](https://github.com/osiam/osiam/pull/149).

PR #149, opened 2016-02-21
Remove attribute removal helper

wallner:
This pull request is considered "work in progress" but it shows a way how we could get rid of the dreadful `AttributeRemovalHelper` at the moment two unit tests have been set to ignore, as they are depending on the `AttributeRemovalHelper` We have to fix them before we can merge this.


dacrome:
I like your solution, because it removes the `AttributeRemovalHelper`, but maybe the attributes filtering can be done in a [custom HTTP message converter](http://stackoverflow.com/questions/30757981/rest-custom-http-message-converter-spring-boot-1-2-3)?

tkrille:
Is this still WiP?

dacrome:
It's funny that the two failing tests are the same: [ShowComplexAttributeFilterSpec](https://github.com/osiam/osiam/pull/149/files#diff-1dfcf1a16b0bdcccd7025f7af682c7e2R43) and [IsoDateFormatSpec](https://github.com/osiam/osiam/pull/149/files#diff-fd9ab2aec04db7d3aaa2f3adad3c6ebfR44). I think you can delete the last one.

tkrille:
They are not the same, they just look like they were :)

The tests are totally useless, because they still base on the old behavior, i.e. serialize the `User`s as JSON string (filtering it thereby), deserialize them back as a `Map` and then let Spring serialize them as part of the ScimSearchResult again. Just delete them @wallner 

The ISODateFormatSpec, for instance, depends on the fact, that the `User` has been serialized before, so that the contained `java.util.Date` is converted to `String` holding the ISO 8601 formatted date.

wallner:
@tkrille I don't think this is work in progress any more. It's ready for review.

dacrome:
Then please change the PR's title

dacrome:
I'm fine with this PR. What about you @tkrille ?

tkrille:
I still couldn't manage to look at it. I'm still on #143. But this one will be next.

src/main/java/org/osiam/resources/controller/GroupController.java
```diff
@@ -117,4 +131,13 @@ public void delete(@PathVariable final String id) {
         headers.setLocation(location);
         return new ResponseEntity<>(group, headers, status);
     }
+
+    private Set<String> extractAttributes(String attributesParameter) {
+        List<String> strings = Lists.newArrayList("schemas");
+        if (!Strings.isNullOrEmpty(attributesParameter)) {
+            strings.addAll(Arrays.asList(attributesParameter.split("[,|\\.]")));
```
tkrille: Why is `.` a valid attribute separator?

src/main/java/org/osiam/resources/controller/GroupController.java
```diff
@@ -117,4 +131,13 @@ public void delete(@PathVariable final String id) {
...
+        }
+        strings.forEach(e -> strings.set(strings.indexOf(e), e.trim()));
```
tkrille: Please move this into the then block. It makes no sense to me, to trim the default attribute (`"schemas"`), if no attributes have been submitted.

src/main/java/org/osiam/resources/controller/GroupController.java
```diff
@@ -117,4 +131,13 @@ public void delete(@PathVariable final String id) {
         headers.setLocation(location);
         return new ResponseEntity<>(group, headers, status);
     }
+
+    private Set<String> extractAttributes(String attributesParameter) {
+        List<String> strings = Lists.newArrayList("schemas");
+        if (!Strings.isNullOrEmpty(attributesParameter)) {
```
tkrille, line 137: You maybe want to invert the condition and turn it into a guard clause.

src/main/java/org/osiam/resources/controller/GroupController.java
```diff
@@ -117,4 +131,13 @@ public void delete(@PathVariable final String id) {
         headers.setLocation(location);
         return new ResponseEntity<>(group, headers, status);
     }
+
+    private Set<String> extractAttributes(String attributesParameter) {
+        List<String> strings = Lists.newArrayList("schemas");
```
tkrille: Why is this a `List` if you convert it to a set in the end anyways? Shouldn't it be a `Set` und just get returned?

src/main/java/org/osiam/resources/controller/GroupController.java
wallner: That is a valid question. The answer is: It isn't. I moved that method here from the `AttributeRemovalHelper`. I would recommend that we leave this here and file a bug and provide a later fix.

tkrille: You could also use `com.google.common.base.Splitter` to split and trim the attributes.

tkrille: Yes, we should stay bug compatible for now.

tkrille: Is it really a bug, though? Must have had some kind of sense, that it was contained in the `ARH` before.

src/main/java/org/osiam/resources/controller/GroupController.java
wallner: I didn't know about `Splitter`

src/main/java/org/osiam/resources/controller/GroupController.java
wallner: The reason is that I wasn't secure enough in my streams api usage. Using the Splitter makes the whole discussion moot. However, I just consulted the scim RFC and we have a bug here: not "schemas" but "id" is supposed to be always returned. I'll file a bug.

src/main/java/org/osiam/resources/scim/Resource.java
```diff
@@ -103,8 +102,9 @@ public Meta getMeta() {
      *
      * @return a the list of schemas as a {@link Set}
      */
-    public Set<String> getSchemas() {
-        return ImmutableSet.copyOf(schemas);
+    @NotEmpty
+    public Set<String> getSchemas(){
+        return schemas;
```
tkrille, line 106: Why this? I mean, it's the SCIM stuff of the server now, so we probably don't need the immutability here, but I still wonder, whether it will break #143.

src/main/java/org/osiam/resources/controller/GroupController.java
```diff
@@ -44,7 +51,11 @@
 import javax.validation.Valid;
 import java.io.IOException;
 import java.net.URI;
+import java.util.Arrays;
```
tkrille: Unused import

src/main/java/org/osiam/resources/controller/GroupController.java
```diff
@@ -44,7 +51,11 @@
...
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
```
tkrille: Unused import

src/main/java/org/osiam/resources/scim/Group.java
```diff
@@ -110,6 +112,7 @@ public String toString() {
                 + getSchemas() + "]";
     }
 
+
```
tkrille: Why this empty line?

src/main/java/org/osiam/resources/scim/User.java
```diff
@@ -523,6 +535,7 @@ public String toString() {
                 + getExternalId() + ", getMeta()=" + getMeta() + ", getSchemas()=" + getSchemas() + "]";
     }
 
+
```
tkrille: Why this empty line?

PR #150, opened 2016-02-21
Refactor configurations

tkrille:
This moves all configurations from org.osiam.Osiam to their dedicated
configuration classes.
See the individual commits for further details.

This time it should actually work and not break everything :)


dacrome:
Thank you!

PR #157, opened 2016-02-25
Change namespace for configuration properties

### LINKED ISSUE #138: Change namespace for configuration properties in YAML config?
After switching to YAML and having OSIAM_HOME in place, I wondered, whether the current namespace for configuration properties is a good choice. Currently we are using `org.osiam.*` for the ordinary properties, but there's also `osiam.home` (or `OSIAM_HOME`) for the special home directory. I really don't like special treatment of such things, so I'd suggest that we change the namespace from `org.osiam.*` to `osiam.*`. That's also more aligned with Spring Boot, because they use `spring.*` and not `org.springframework.spring-boot.*` (or whatever). I don't see any chance of a naming collision either. 

What do you think?

dacrome:
Resolves #138

### DIFF
--- a/docs/detailed-reference-installation.md
+++ b/docs/detailed-reference-installation.md
@@ -338,7 +338,7 @@ First you have to set the LDAP configuration in the [osiam.yaml]
 
 The available properties are:
 
-**org.osiam.ldap.enabled**
+**osiam.ldap.enabled**
 
 Default = false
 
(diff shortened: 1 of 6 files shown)


dacrome:
Corresponding IT change: https://github.com/osiam/connector4java-integration-tests/pull/328

wallner:
Looks good, but please rebase.

dacrome:
Rebased

wallner:
Thank you.

PR #163, opened 2016-02-28
Don't explicitly set sources in Servlet environment

tkrille:
The current class inheriting `SpringBootServletInitializer` will be used
as application sources, if none is defined.


dacrome:
Thanks!

PR #164, opened 2016-02-28
Initialize home directory via command line

tkrille:
This adds a command to OSIAM, that allows one to initialize the home
directory without starting OSIAM.
The command parser is dead-simple and should be replaced by something
more sophisticated in the future, if we want to support more cli
commands.
The default command is `server` so one don't has to specify it
every time.
Depending on the command, the Application Context will be bootstrapped
with a different `@Configuration` class.
The `InitHome` class represents the simplest (but most sufficient)
configuration to initialize the home directory.
Use a customized `@ComponentScan` definition, to prevent loading the
`InitHome` bean during normal server startup.

### DIFF
--- a/src/main/java/org/osiam/Osiam.java
+++ b/src/main/java/org/osiam/Osiam.java
@@ -25,21 +25,28 @@ package org.osiam;
 
 import com.google.common.collect.ImmutableMap;
 import com.ryantenney.metrics.spring.config.annotation.EnableMetrics;
+import org.osiam.cli.InitHome;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.boot.builder.SpringApplicationBuilder;
 import org.springframework.boot.context.web.SpringBootServletInitializer;
 import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.ComponentScan;
 import org.springframework.context.annotation.EnableAspectJAutoProxy;
+import org.springframework.context.annotation.FilterType;
 import org.springframework.security.authentication.encoding.ShaPasswordEncoder;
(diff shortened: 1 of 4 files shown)


src/main/java/org/osiam/Osiam.java
```diff
@@ -48,11 +55,32 @@
...
+            application.setSources(Collections.<Object>singleton(InitHome.class));
+            application.setWebEnvironment(false);
+        } else {
+            application.setSources(Collections.<Object>singleton(Osiam.class));
+        }
+
         application.run(args);
     }
 
+    private static String extractCommand(String[] args) {
```
dacrome: `args` is a common abbreviation. But maybe in this function you could avoid using `arg` or `*Arg` and using `argument` instead?

dacrome:
Works like a charm. After you checked [codacy](https://www.codacy.com/app/OSIAM/osiam/pullRequest?prid=191092) and we discussed [this](https://github.com/osiam/osiam/pull/164#discussion_r54344723), I will merge your PR.

tkrille:
Updated

src/main/java/org/osiam/OsiamHome.java
```diff
@@ -33,11 +33,7 @@
 import org.springframework.stereotype.Component;
 
 import java.io.IOException;
-import java.nio.file.DirectoryStream;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.nio.file.Paths;
-import java.nio.file.StandardCopyOption;
+import java.nio.file.*;
```
dacrome: Please disable wildcard imports in your IDE

tkrille: I couldn't care less than about imports. But whatever ... I'll fix it. Give me one moment, please.

tkrille:
Updated

dacrome:
Thanks a lot!

PR #165, opened 2016-02-28
Remove auth-server client from merged server

### LINKED ISSUE #49: Remove auth-server client from merged server
After merging the core servers, the auth-server client is not necessary anymore. Should be automatically removed by a Flyway migration.

tkrille:
Rename the migration class, because from now on it's responsibility is
not only dealing with the scopes, but also to do all data migrations
from 2.5 to 3.0.

Resolves #49


wallner:
Thank you!

PR #169, opened 2016-02-28
Remove creation of example data via Flyway

### LINKED ISSUE #45: Remove creation of example data via Flyway
As discussed in #43, we want to remove the creation of example data via Flyway. The first client, user and group, etc. should be created programmatically on the "first start", e.g. when OSIAM starts and the database is empty. Obviously needs #44 to be merged first.

tkrille:
Example data will be created by a special class.
It checks whether the database already contains data and if not,
it creates the example data on startup.

Resolves #45

### DIFF
--- a/src/main/java/org/osiam/ExampleData.java
+++ b/src/main/java/org/osiam/ExampleData.java
@@ -1,0 +1,80 @@
+/**
+ * The MIT License (MIT)
+ *
+ * Copyright (C) 2013-2016 tarent solutions GmbH
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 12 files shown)


src/main/java/org/osiam/ExampleData.java
```diff
@@ -0,0 +1,80 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import javax.annotation.PostConstruct;
+
+@Component
+public class ExampleData {
+
+    public static final Logger logger = LoggerFactory.getLogger(ExampleData.class);
```
dacrome: private

dacrome:
Please update changelog

tkrille:
Updated

dacrome:
Thank you!

PR #170, opened 2016-02-28
Always return the `id` attribute

### LINKED ISSUE #151: Returning the "id" attribute is mandatory
According to the [SCIM specification](https://tools.ietf.org/html/rfc7643#section-7) the attribute "id" is required and must always be returned. At the moment we are always returning the "schemas" attribute.

tkrille:
The `id` attribute is defined with "a 'returned' characteristic of
always" [1].

[1] https://tools.ietf.org/html/rfc7643#section-3.1

Fixes #151

### DIFF
--- a/src/main/java/org/osiam/resources/controller/GroupController.java
+++ b/src/main/java/org/osiam/resources/controller/GroupController.java
@@ -143,6 +143,7 @@ public class GroupController {
             );
         }
         result.add("schemas");
+        result.add("id");
         return result;
     }
 }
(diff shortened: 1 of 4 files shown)


wallner:
Are you adding tests too?

tkrille:
Only if I have to :)

tkrille:
If you tell me how, @wallner, I can do this. Unfortunately, there are no tests at all, that care about the attribute filtering when searching for resources. I really don't know how to do this with MockMVC and the current attribute filtering process. As you are the author of both, please give me some advice how to test the stuff in this PR.

tkrille:
Alright, I've added 2 simple tests.

wallner:
Excellent. Thank you!

Issue #51, opened 2015-10-23
Check the changelog


tkrille:
There might be changes in auth- and resource-server, that have not been added to the respective changelogs. As the merged changelog is a combination of the aforementioned changelogs, we should check the latest changes in the auth- and resource-server and extend the merged changelog if necessary.

PR #172, opened 2016-03-02
Refactor processing of OSIAM's home directory

tkrille:
OSIAM will now load the configuration only from the YAML file in its
home directory. This is done by running the initialization as early as
possible via an `ApplicationListener`. Initialization will take place
when the environment is ready to use, but before external configuration
will be loaded by the `ConfigFileApplicationListener`.

OSIAM will now load the configuration only from the YAML file in its
home directory. This is done by running the initialization as early as
possible via an `ApplicationListener`. Initialization will take place
when the environment is ready to use, but before external configuration
will be loaded by the `ConfigFileApplicationListener`.

OSIAM will now load the configuration only from the YAML file in its
home directory. This is done by running the initialization as early as
possible via an `ApplicationListener`. Initialization will take place
when the environment is ready to use, but before external configuration
will be loaded by the `ConfigFileApplicationListener`.

### DIFF
--- a/src/main/java/org/osiam/Osiam.java
+++ b/src/main/java/org/osiam/Osiam.java
@@ -51,14 +51,11 @@ import java.util.Map;
 public class Osiam extends SpringBootServletInitializer {
 
     private static final Map<String, Object> DEFAULT_PROPERTIES = ImmutableMap.<String, Object>of(
-            "osiam.home", System.getenv("HOME") + "/.osiam",
-            "spring.config.location", "classpath:/home/config/osiam.yaml,file:${osiam.home}/config/osiam.yaml"
+            "spring.config.location", "file:${osiam.home}/config/osiam.yaml"
     );
 
     public static void main(String[] args) {
         SpringApplication application = new SpringApplication();
-        application.setDefaultProperties(DEFAULT_PROPERTIES);
-
         String command = extractCommand(args);
         if ("initHome".equals(command)) {
             application.setSources(Collections.<Object>singleton(InitHome.class));
@@ -66,7 +63,8 @@ public class Osiam extends SpringBootServletInitializer {
         } else {
             application.setSources(Collections.<Object>singleton(Osiam.class));
         }
-
+        application.addListeners(new OsiamHome());
+        application.setDefaultProperties(DEFAULT_PROPERTIES);
         application.run(args);
     }
 
@@ -84,6 +82,7 @@ public class Osiam extends SpringBootServletInitializer {
     @Override
     protected SpringApplicationBuilder configure(SpringApplicationBuilder applicationBuilder) {
         applicationBuilder.application().setDefaultProperties(DEFAULT_PROPERTIES);
+        applicationBuilder.application().addListeners(new OsiamHome());
         return applicationBuilder;
     }
 
--- a/src/main/java/org/osiam/OsiamHome.java
+++ b/src/main/java/org/osiam/OsiamHome.java
@@ -26,11 +26,16 @@ package org.osiam;
 import com.google.common.base.Strings;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.context.config.ConfigFileApplicationListener;
+import org.springframework.boot.context.event.ApplicationEnvironmentPreparedEvent;
+import org.springframework.boot.context.event.ApplicationPreparedEvent;
+import org.springframework.context.ApplicationEvent;
+import org.springframework.context.ApplicationListener;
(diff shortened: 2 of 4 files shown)


wallner:
Two missing 'be' in the PR's description, also 'has not yet been set up'

tkrille:
Thanks for your comments. I fixed the commit message.

src/main/java/org/osiam/Osiam.java
```diff
@@ -51,22 +51,20 @@
...
-            "spring.config.location", "classpath:/home/config/osiam.yaml,file:${osiam.home}/config/osiam.yaml"
+            "spring.config.location", "file:${osiam.home}/config/osiam.yaml"
     );
 
     public static void main(String[] args) {
         SpringApplication application = new SpringApplication();
-        application.setDefaultProperties(DEFAULT_PROPERTIES);
-
         String command = extractCommand(args);
         if ("initHome".equals(command)) {
```
wallner, line 60: I'm not a huge fan of Yoda notation.

dacrome, line 60: But codacy ;)

src/main/java/org/osiam/OsiamHome.java
```diff
@@ -41,57 +46,72 @@
...
-    public String getI18nDirectory() {
-        return osiamHomeDir.resolve("i18n").toUri().toString();
+    public void configure(ConfigurableEnvironment environment) {
+        String rawOsiamHome = environment.getProperty("osiam.home", System.getenv("HOME") + "/.osiam");
+        checkState(!Strings.isNullOrEmpty(rawOsiamHome), "'osiam.home' is not set");
+        osiamHome = Paths.get(rawOsiamHome).toAbsolutePath();
+        environment.getPropertySources().addFirst(new PropertySource<Path>("osiamHome", osiamHome) {
+            @Override
+            public Object getProperty(String name) {
+                return "osiam.home".equals(name) ? source.toString() : null;
```
wallner, line 75: Really, Yoda notation?

src/main/java/org/osiam/Osiam.java
tkrille, line 60: Wise to use it is :)

src/main/java/org/osiam/OsiamHome.java
tkrille, line 75: See https://github.com/osiam/osiam/pull/172#discussion_r54946958 :)

src/main/java/org/osiam/Osiam.java
tkrille, line 60: But seriously, I think it's ok to use it. I know, that, when reversed, it might be a little bit easier to comprehend. But OTOH it's a well-known idiom in the Java community. So I think there are not really arguments against nor for it.

wallner:
I'm not completely sure if you intended it to work, if not please remove it from the PR, but `java -jar .osiam.war initHome` doesn't work for me. It exits with "java.lang.IllegalArgumentException: Could not resolve placeholder 'osiam.home' in string value "file:${osiam.home}/config/osiam.yaml"

src/main/java/org/osiam/Osiam.java
wallner, line 60: I'm ok with accepting it in your PRs as long as you don't force it on me.

dacrome:
@wallner you need to set osiam.home like `$  java -jar target/osiam.war initHome --osiam.home=/Users/david/.osiam`

wallner:
@dacrome Then it would work, yes. However I'd expect it to create the directory in my home directory if I ommited `--osiam.home=...` at least that would be the principle of least surprise.

wallner:
Can you fix the weird line breaks for the commit message as well, please?

src/main/java/org/osiam/Osiam.java
tkrille, line 60: I would never do such things to you @wallner :)

tkrille:
Sorry for the inconvenience @wallner. Shame on me, I did not test this use case :( I fixed it. To fix it I had to empty the class `org.osiam.cli.InitHome`. See the comment in the class for an explanation.

The line breaks are at the recommended line width of 72 characters. Yes, they're weird sometimes, I must admit. I really don't want to start a discussion about rules for commit message formatting, do you?

tkrille:
Oh, I see what you mean. Fixed it.

wallner:
Thank you. It works lie a charm now and was no inconvenience at all, that's what code reviews are for. Please add a CHANGELOG entry for this change and then it's good for a merge.

tkrille:
There will be no changelog entry for this, because it's a refactoring of something that does not even exist in any released OSIAM version. And apparently it's not even completely finished yet. So what am I supposed to write in the Changelog? We also still have #51, which deals with all this stuff.

wallner:
Ok fair enough. I'm happy to merge then.

PR #173, opened 2016-03-03
Set minimum required maven version to 3.2.1

dacrome:


tkrille:
Do you really want to mix 2 totally unrelated changes?

wallner:
Please either turn that into two separate requests or adjust the commit message accordingly

dacrome:
@wallner and @tkrille I removed the test dependency version bump.

wallner:
Thank you.

PR #175, opened 2016-03-05
Update docs to use PostgreSQL 9.4

tkrille:
Also remove a 'package-info' file, as the information it contains is
outdated.


dacrome:
Thank you

PR #178, opened 2016-03-06
Initialize database via command line

### LINKED ISSUE #167: Add command to run database migration/initialization without starting OSIAM
This has the same rationale as the `initHome` command. It's mostly useful for automated installs, that have to add custom extensions or clients to the database before starting to use OSIAM. In such a scenario it should not be necessary to start OSIAM first. I think we should add a command `initDb` or `migrateDb` that will run the Flyway migrations against the configured database. I'd like to see this in 3.0, because it seems easy to do and gives the users great value.

tkrille:
This adds a command to OSIAM, that allows one to initialize the home
directory without starting OSIAM. The home directory must already be
initialized and any configuration has to be done, before this command
can be used. This should prevent accidental faulty database migrations,
because the command fails, if no configuration can be found in the home
directory. Initialization will be switched off in `OsiamHome` to realize
this.

Resolves #167

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@
 - Support for H2 database has been added and a file-based one is the default
   configuration from now on. The usage scenarios are small installations,
   testing and development.
+- Migration and initialization of the database can be done from the command line
+  with the `migrateDb` command. See [Initialize the Database from the Command Line]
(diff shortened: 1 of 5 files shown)


dacrome:
Thank you!

PR #183, opened 2016-03-07
Remove migration of client scopes

### LINKED ISSUE #166: Don't mess around with the scopes of the add-ons' OAuth clients in database migration
I think we should remove the [migration of the add-ons' clients' scopes](https://github.com/osiam/osiam/blob/master/src/main/java/db/migration/V2__replace_old_scopes.java#L51). In the latest version of the add-ons we've already dropped support for the old scopes, so users had to do the migration anyway by now. As we will only support a migration from the latest version of OSIAM and all it's components, we can safely remove the Flyway migration.

tkrille:
In the latest version of the add-ons we've already dropped support for
the method-based scopes. This way, users had to do the migration anyway
by now. As we will only support a migration from the latest version of
OSIAM and all it's components, remove the Flyway migration.

Furthermore we don't know how users utilize the "example-client", so
removing the scopes and adding both "ADMIN" and "ME" could break their
setup or even make their OSIAM insecure.

Resolves #166

### DIFF
--- a/src/main/java/db/migration/V2__migrations_from_2_5.java
+++ b/src/main/java/db/migration/V2__migrations_from_2_5.java
@@ -47,12 +47,6 @@ public class V2__migrations_from_2_5 implements JdbcMigration, MigrationChecksum
     public void migrate(Connection connection) throws Exception {
         this.connection = connection;
         removeClient("auth-server");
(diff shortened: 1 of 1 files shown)


dacrome:
Thanks!

PR #189, opened 2016-03-12
Use H2 automatic mixed mode by default

tkrille:
This will start an embedded H2 server on a random port. Only clients,
that have access to the database file, can connect to this server. This
has the advantage, that you can connect to the H2 database while OSIAM
is running, e.g. for backups. See [1] for details.

[1] http://www.h2database.com/html/features.html#auto_mixed_mode


dacrome:
Thank you!

PR #191, opened 2016-03-13
Create a fully executable WAR

tkrille:
This should partly fix the issue on how one should run OSIAM in
production. A fully executable WAR can be easily started on a command
line. It also supports fundamental SysVinit operations, like start,
stop. See [1] for more details.

[1] http://docs.spring.io/spring-boot/docs/current/reference/html/deployment-install.html#deployment-install


dacrome:
Works like a charm! Thank you!

PR #193, opened 2016-03-14
Make log level configurable

tkrille:
The configuration property is called `osiam.logging.level`, because of
the decision to wrap all Spring Boot properties [1]. It also creates a
new namespace, which leaves enough space for future extensions. The
property is bound to the actual Spring Boot property by using the same
technique as described in "Use ‘short’ command line arguments" [2] in
the Spring Boot manual.

[1] osiam#35
[2] https://docs.spring.io/spring-boot/docs/current/reference/html/howto-properties-and-configuration.html#howto-use-short-command-line-arguments


wallner:
Nice. Thank you!

Issue #153, opened 2016-02-24
Drop support for fake Facebook API


tkrille:
It seems that nobody uses this feature. We also want to change the `/Me` endpoint (#141) to a SCIM conform implementation. So the idea is to drop the whole Facebook support for now. If somebody really needs this, we can add it again later.

tkrille:
There are still some residues left: https://github.com/osiam/osiam/blob/a5d9e00afaca9310b686c3df1ce15fee894c1934/src/main/java/org/osiam/configuration/FacebookClientCredentialsSecurity.java and probably more.

PR #198, opened 2016-03-25
Fix '/Me' endpoint

wallner:
This commit fixes the '/Me' endpoint by returning a fully populated
`User`-Object. It also removes the last residues of the ill devised
facebook api.


src/main/java/org/osiam/resources/package-info.java
```diff
@@ -35,8 +35,6 @@
  *  <p>/Group - org.osiam.resources.controller.GroupController - is a Controller to create, replace, modify, get,
  *  delete and search groups.</p>
  *
- *  <p>/me - org.osiam.resources.controller.MeController - is a Facebook /me clone to get name, email and id of an user.</p>
```
dacrome, line 38: Why not replace the description?

dacrome:
Please add `Closes #153` to the PR description

src/main/java/org/osiam/resources/package-info.java
wallner, line 38: Because we are planning to get rid of the `package-info.java` files in the servers anyways. There is now inherent value in them.

wallner:
Updated.

dacrome:
Thank you!

PR #201, opened 2016-03-29
Add the generated-sources/apt directory as a source directory.

### LINKED ISSUE #200: generated-sources/apt not being detected as a source directory when importing into eclipse
The hibernate JPA DDL  generated sources are not added to an eclipse project when importing. This is caused because there is no configuration in the POM to add them. It's easy to reproduce, just import the project using the POM in eclipse and notice all of the Entity_ classes cannot be found.

A fix would be to use the build-helper-maven-plugin so eclipse is aware of the sources. I'll submit a pull request later in case you're interested in just doing that.

tkent:
This will cause eclipse to recognize the Hibernate JPA DDL generated sources when the project is imported.

Fixes #200


dacrome:
Hey @tkent. Thanks for your PR.

I usually use IntelliJ, but I downloaded Eclipse to test your change and it's not working or only if I run maven on command line before. In IntelliJ the ["Annotation Processors" needs to be disabled](https://www.jetbrains.com/help/idea/2016.1/configuring-annotation-processing.html) to use OSIAM without any errors. Do I need to change any preferences in Eclipse?

tkent:
Hey @dacrome,

I never actually tried to do an import without doing a compile first. After testing locally, there doesn't seem to be an easy way to coax the m2e plugin into generating the sources before importing. So the fix I included doesn't get you to a seamless import scenario with eclipse.

I ended up finding that the Eclipse setting for annotation processing both generates the sources on import _and_ adds the correct source folder, so I'm going to close this PR request.

For reference, the setting is Preferences -> Maven -> Annotation Processing, then select "Automatically configure JDT APT"

PR #203, opened 2016-04-01
Persist the Multi-Valued display sub-attribute

### LINKED ISSUE #81: 'Display value' not saved
When editing a user, the field `Display value` of each type of multi-valued attribute will not be saved.

Steps to reproduce:

1. Edit a user
2. For each multi-valued attribute do:
  1. Add 1 element
  2. Type something in the field `Display value'
3. Save the user
4. Open the edit form again

Expected:

Each multi-valued attribute has its field `Display value` set with the text entered in step 2.

Actual:

The field `Display value` of each multi-valued attribute is empty.

### LINKED ISSUE #199: Display value of the multi-valued attributes is not persisted
While I wanted to fix the https://github.com/osiam/addon-administration/issues/81 bug in the administration it turns out that the display value of all multi value attributes (like email or phone number) is not persisted, because there is not even a column in the database.

dacrome:
Resolves #199 and closes osiam/addon-administration#81

### DIFF
--- a/src/main/java/db/migration/V2__migrations_from_2_5.java
+++ b/src/main/java/db/migration/V2__migrations_from_2_5.java
@@ -47,6 +47,49 @@ public class V2__migrations_from_2_5 implements JdbcMigration, MigrationChecksum
     public void migrate(Connection connection) throws Exception {
         this.connection = connection;
         removeClient("auth-server");
+        addDisplayToMultiValuedAttributes();
+    }
+
+    private void addDisplayToMultiValuedAttributes() throws SQLException {
+        try (PreparedStatement statement = connection
+                .prepareStatement("ALTER TABLE scim_address ADD display VARCHAR(255)")) {
+            statement.execute();
+        }
+
+        try (PreparedStatement statement = connection
+                .prepareStatement("ALTER TABLE scim_certificate ADD display VARCHAR(255)")) {
+            statement.execute();
+        }
+
+        try (PreparedStatement statement = connection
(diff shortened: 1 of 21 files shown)


tkrille:
This will be done after OSIAM 3.0

dacrome:
Why not with 3.0?

tkrille:
Mostly, because yesterday was the last day to add new features or changes. Of course, that's negotiable. What do you think @wallner?

src/main/java/db/migration/V2__migrations_from_2_5.java
```diff
@@ -47,6 +47,49 @@ public Integer getChecksum() {
     public void migrate(Connection connection) throws Exception {
         this.connection = connection;
         removeClient("auth-server");
+        addDisplayToMultiValuedAttributes();
+    }
+
+    private void addDisplayToMultiValuedAttributes() throws SQLException {
```
tkrille, line 53: Maybe this should be done with an SQL-based migration?

tkrille:
Alright, I didn't notice, that you already created an issue for that more than a week ago. Rescheduled for 3.0.

src/main/java/db/migration/V2__migrations_from_2_5.java
dacrome, line 53: But then I have to do it for every database.

tkrille, line 53: Alright. Maybe we should migrate to a Spring JDBC migration, to make this less complex, but that should be done in an upcoming PR.

tkrille:
LGTM

PR #204, opened 2016-04-01
Simplify command line parser

tkrille:
The former version was a workaround, so one could use the Spring Boot
Maven plugin, because it only ever passes one argument to the
application, e.g. `initHome --osiam.home=...`. Since Spring Boot 1.3
this workaround doesn't apply anymore, so it can be removed. It wasn't
even safe, because it could fail on some edge cases.

Use a Java 8 stream in the hope that it's easier to comprehend.


dacrome:
Thanks!

PR #205, opened 2016-04-02
Create OSIAM extension programmatically

tkrille:
Doing it in the database migration is not a good idea. Due to the fact that each migration has a checksum, we would not be able to change what should initially be created. Example data has already been migrated to programmatic creation too.

This also contains a refactoring of the extension DAO. See the corresponding commit for details.


src/test/groovy/org/osiam/scim/extension/OsiamExtensionTest.groovy
```diff
@@ -0,0 +1,63 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package org.osiam.scim.extension
+
+import org.osiam.resources.scim.ExtensionFieldType
+import org.osiam.storage.ExtensionRepository
+import spock.lang.Specification
+
+class OsiamExtensionTest extends Specification {
```
dacrome: `OsiamExtensionTest`? I guess `OsiamExtensionSpec`.

tkrille: Yes, apparently I hit the wrong buttons in IntelliJ :)

tkrille:
Updated

dacrome:
Thank you!

PR #209, opened 2016-04-10
Make extensions configurable

### LINKED ISSUE #105: Document that extensions do not support complex types
I can't see any documentation of this fact, but it has to be documented somewhere.

### LINKED ISSUE #186: Make extensions configurable
Currently extensions and extension fields can only be created by sql. It would be a great feature to create extensions by configuration (yaml). The extensions can only be created, any change needs to be done by sql. When the extension does not exists on startup, it will be created.

tkrille:
Read extensions from `osiam.yaml` and create them on start up. If the
extension already exists, don't create it. Add new value classes for
extension and extension field, because the `ExtensionEntity` stores the
fields as `Set` and Spring is unable to bind a YAML list to a `Set`.
This is due to lists are ordered, but sets not. Use Bean validation to
validate the configured extensions, which keeps the code simpler.

Closes #186
Closes #105

### DIFF
--- a/docs/detailed-reference-installation.md
+++ b/docs/detailed-reference-installation.md
@@ -60,7 +60,7 @@ The `.war` file is actually an executable application.
 This makes it possible to run it on the command line like any other command.
 So, instead of starting it with `java -jar osiam.war` you can just run `./osiam.war`.
 
-**Note:** This probably works on UNIX and similar systems (like Linux or OSX) only.
+**NOTE:** This probably works on UNIX and similar systems (like Linux or OSX) only.
 It might work in a cygwin environment, but that has not been thoroughly tested.
 Essentially you need Bash, some standard UNIX commands and Java.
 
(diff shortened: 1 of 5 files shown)


dacrome:
Please change the [configuration in the ITs](https://github.com/osiam/connector4java-integration-tests/tree/master/src/test/docker/osiam) just to make sure OSIAM is starting

docs/detailed-reference-installation.md
```diff
@@ -381,50 +381,65 @@ Now you are ready to work with OSIAM, either using a [connector]
 
 ### Configuring SCIM Extension
 
-At the moment you can register an extension and all users will have this
-additional self-defined fields. There are no pure SCIM core schema users
-anymore. The additional fields could be empty, so it will be optional for the
-users to fill them out with values.
+You can extend OSIAM's data model with custom extension.
```
wallner: "...with custom extensions." A typo lost the period at the end of the sentence.

docs/detailed-reference-installation.md
```diff
@@ -381,50 +381,65 @@ Now you are ready to work with OSIAM, either using a [connector]
...
 
-At the moment you can register an extension and all users will have this
-additional self-defined fields. There are no pure SCIM core schema users
-anymore. The additional fields could be empty, so it will be optional for the
-users to fill them out with values.
+You can extend OSIAM's data model with custom extension.
+This is a feature of [SCIM 2](http://www.simplecloud.info/), the data model and API that OSIAM implements.
+Extensions add new attributes to resources using a unique namespace.
+These attributes can be mandatory or optional and have a fixed data type.
+See [the section about SCIM 2 extensions](https://tools.ietf.org/html/rfc7643#section-3.3) for more information.
```
wallner: Maybe "see the SCIM 2 RFC section about extensions for more information."

docs/detailed-reference-installation.md
```diff
@@ -381,50 +381,65 @@ Now you are ready to work with OSIAM, either using a [connector]
...
-  <SELFDEFINED_FIELD_NAME>,
-  <TYPE>,
-  <UNIQUE_EXTENSION_ID>)`
-```
+**NOTE:** OSIAM does not yet support complex data types for extensions.
+The absence of this feature will be addressed in a future version.
 
-**NOTE:** You may leave the UNIQUE_EXT_FIELD_ID off and let your database
-generate a unique id, for example:
+See [the section about SCIM 2 data types](https://tools.ietf.org/html/rfc7643#section-2.3) for more information.
```
wallner: Maybe "see the SCIM 2 RFC section about data types for more information."

docs/detailed-reference-installation.md
```diff
@@ -381,50 +381,65 @@ Now you are ready to work with OSIAM, either using a [connector]
...
 
-`INSERT INTO scim_extension_field VALUES(213, false, 'gender', 'STRING', 39)`
+OSIAM creates all configured extensions on startup, only if they do not exist in the database.
```
wallner: I don't think the comma is right.

docs/detailed-reference-installation.md
```diff
@@ -381,50 +381,65 @@ Now you are ready to work with OSIAM, either using a [connector]
...
+It will only check, if an extension with the same URN is already defined or not.
```
wallner: It will only check it an extension with the same URN is already defined.

src/main/java/org/osiam/scim/extension/ExtensionsConfiguration.java
```diff
@@ -0,0 +1,139 @@
...
+import java.util.List;
+
+import static java.util.stream.Collectors.toSet;
+
+@Component
+@ConfigurationProperties("osiam.scim")
+public class ExtensionsConfiguration {
+
+    private final ExtensionRepository extensionRepository;
+    private final List<Extension> extensions = new ArrayList<>();
```
wallner: Isn't a `Set` more appropriate syntactically? After all we are should not allow the configuration of the same extension twice?

src/main/java/org/osiam/scim/extension/ExtensionsConfiguration.java
```diff
@@ -0,0 +1,139 @@
...
+    @PostConstruct
+    public void createExtensions() {
+        extensions.stream()
+                .filter(extension -> !extensionRepository.existsByUrnIgnoreCase(extension.getUrn()))
+                .map(Extension::toEntity)
+                .forEach(extensionRepository::saveAndFlush);
+    }
+
+    @Valid
+    public List<Extension> getExtensions() {
```
wallner, line 64: Is public visibility a good idea? We are not handling the `ExtensionConfiguration.Extension` anywhere else then in this class, so maybe we could consider private visibility. Or at least package wide default visibility, to keep the Spock test happy in the IDE, although it shouldn't really care. I'm not even sure we need to expose this method to be honest, all we want to do is create new extensions in the database, right?

src/main/java/org/osiam/scim/extension/ExtensionsConfiguration.java
```diff
@@ -0,0 +1,139 @@
...
+        return extensions;
+    }
+
+    public static class Extension {
```
wallner, line 68: See the notes about visibility above.

src/main/java/org/osiam/scim/extension/ExtensionsConfiguration.java
```diff
@@ -0,0 +1,139 @@
...
+        public void setUrn(String urn) {
+            this.urn = urn;
+        }
+
+        public List<Field> getFields() {
+            return fields;
+        }
+    }
+
+    public static class Field {
```
wallner, line 97: See the notes about visibility above.

src/main/resources/home/config/osiam.yaml
```diff
@@ -66,6 +66,25 @@ osiam:
     #connection-timeout-ms: 30000
 
   #
+  # SCIM Extensions
+  #
+  scim:
+    extensions:
```
wallner, line 72: Can we come up with nicer examples?

wallner:
Looks good, please take a look at my comments.

src/main/java/org/osiam/scim/extension/ExtensionsConfiguration.java
tkrille, line 64: It has to be public, because of Spring's data binding. `@ConfigurationProperties` requires the annotated class to be more or less a Java bean. See also: https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-external-config.html#boot-features-external-config-typesafe-configuration-properties

docs/detailed-reference-installation.md
tkrille: I'll remove the `only`, as it's just line noise.

docs/detailed-reference-installation.md
tkrille: I'll reword it.

docs/detailed-reference-installation.md
tkrille: I'll reword it

docs/detailed-reference-installation.md
tkrille: Done.

docs/detailed-reference-installation.md
tkrille: Done

src/main/resources/home/config/osiam.yaml
tkrille, line 72: Any suggestions?

tkrille:
Thanks for the suggestion @dacrome. I'll do it right away.

tkrille:
Thanks for your comments @wallner. I've updated the PR and answered your questions.

tkrille:
Here you are, @dacrome: https://github.com/osiam/connector4java-integration-tests/pull/338

wallner:
Thank you

PR #210, opened 2016-04-10
Fix JPA model generation error in IntelliJ IDEA

tkrille:
IDEA only picks up the Jar, if it is on the classpath. Use an optional
dependency in the POM to realize this. The version is already defined by
Spring Boot's parent POM.


dacrome:
Thank you!

PR #212, opened 2016-04-16
Disable banner

### LINKED ISSUE #208: Remove Spring Boot Banner
I know, the Spring guys think it's cool, and we should use our own application's banner, but it gains us nothing and causes problems with log shippers. I propose that we turn off the banner, starting with 3.0.

tkrille:
The banner gains us nothing and might cause problems for log shippers.

Closes #208

### DIFF
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -32,6 +32,7 @@ logging.level:
   org.springframework: WARN
   org.osiam: ${osiam.logging.level:INFO}
 
+spring.main.banner-mode: 'off'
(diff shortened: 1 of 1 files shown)


dacrome:
What about `application.setBannerMode(Banner.Mode.OFF)`?

tkrille:
Because then I would have to set it 2 times: for standalone and WAR deployment.

tkrille:
Also, this way users can turn it on again, if they like. Don't know why somebody would want to do so, but ... :)

dacrome:
Ok. Thank you!

PR #213, opened 2016-04-16
Fix error when using a wrong username

tkrille:
User provisioning doesn't return null, but throws an exception, if the
username could not be found.

Part of #48


dacrome:
Thanks!

PR #214, opened 2016-04-16
Replace `for` loop with Java 8 streams

tkrille:
It's more idiomatic, shorter and easier to understand.


wallner:
Thank you. We should probably do this once for the whole project.

tkrille:
Yes, definitively. I just stumbled across this one as I prepared another PR.

PR #215, opened 2016-04-16
Cleanup exception handling in provisioning services

tkrille:
The removed `try-catch` blocks are handling exceptions, that will never
bubble up to them. They also hide sever errors of the persistence layer,
by handling them as "resource not found" errors.


wallner:
It's way better this way. Thank you!

PR #220, opened 2016-05-10
Bump Spring Boot version

dacrome:
Also the aspectj versions and the objenesis version


tkrille:
I think we can ignore the warning from Spring at this moment. Seems like something changed in Spring Core and Spring Security OAuth has not updated yet. In my opinion, it does not expose a security risk. So, thanks for your work!

PR #225, opened 2016-06-06
Assign controllers to the right exception handlers

### LINKED ISSUE #48: Check exception handling

tkrille:
`LoginController` and `AccessConfirmationController` are completely
serving the web interfaces, while `TokenController` is part of the REST
API. The package `org.osiam.security.controller` only contains these 3
controllers.

Resolves #48

### DIFF
--- a/src/main/java/org/osiam/auth/exception/WebExceptionHandler.java
+++ b/src/main/java/org/osiam/auth/exception/WebExceptionHandler.java
@@ -23,6 +23,8 @@
  */
 package org.osiam.auth.exception;
 
+import org.osiam.security.controller.AccessConfirmationController;
+import org.osiam.security.controller.LoginController;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


dacrome:
Thanks!

PR #226, opened 2016-06-07
Publish latest snapshot as `osiam-latest` to bintray

dacrome:
Because it is not necessary to publish different snapshot as
maven artifact, this commit change it to only one available
artifact on bintray:
https://dl.bintray.com/osiam/OSIAM/osiam-latest.war


dacrome:
I uploaded it directly in the root folder of all OSIAM repositories. I can upload to https://dl.bintray.com/osiam/OSIAM/org/osiam/osiam/ but this folder is a maven repo so I'm not sure if it is a good idea.

tkrille:
To do this right, we actually have to create a new "Generic" Bintray repository. But this can be done later. What matters is, that we have the latest snapshot accessible under a static URL.

pom.xml
```diff
@@ -86,11 +86,6 @@
             <name>osiam-releases</name>
             <url>https://api.bintray.com/maven/osiam/OSIAM/org.osiam:osiam/;publish=1</url>
         </repository>
-        <snapshotRepository>
-            <id>snapshots</id>
-            <name>osiam-snapshots</name>
-            <url>https://oss.jfrog.org/artifactory/oss-snapshot-local</url>
-        </snapshotRepository>
```
tkrille, line 93: Removing this leads to uploading the snapshot to Bintray releases repo:

```
Uploading: https://api.bintray.com/maven/osiam/OSIAM/org.osiam:osiam/;publish=1/org/osiam/osiam/3.0-SNAPSHOT/osiam-3.0-20160608.081458-1.war
```

I don't think that this is the intended behavior.

tkrille, line 93: Also, the build fails:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-deploy-plugin:2.8.2:deploy (default-deploy) on project osiam: Failed to deploy artifacts: Could not transfer artifact org.osiam:osiam:war:3.0-20160608.081458-1 from/to releases (https://api.bintray.com/maven/osiam/OSIAM/org.osiam:osiam/;publish=1): Failed to transfer file: https://api.bintray.com/maven/osiam/OSIAM/org.osiam:osiam/;publish=1/org/osiam/osiam/3.0-SNAPSHOT/osiam-3.0-20160608.081458-1.war. Return code is: 400, ReasonPhrase: Bad Request. -> [Help 1]
```

Maybe we have to disable the deploy plugin for snapshot builds?

dacrome:
> To do this right, we actually have to create a new "Generic" Bintray repository. But this can be done later. What matters is, that we have the latest snapshot accessible under a static URL.

But when we do this later, then the URL changes.
The "generic" Bintray repo is for all artifacts (osiam and addons-*) which are not need to be accessible as maven dependencies?

dacrome:
I close this PR to create a new one with some changes with discussed offline, like to dual deployment to bintray and jfrog to satisfy the integration-tests and the different download link etc.

PR #231, opened 2016-06-08
Change URLs in docs to the standalone deployment ones

### LINKED ISSUE #224: Examples in Docs are using the wrong URLs
The documentation is using example URLs that reflect the deployment of OSIAM in a Tomcat Servlet container. I think this must be changed to the standalone deployment URLs, e.g.:

`http://localhost:8080/osiam/ServiceProviderConfig` -> `http://localhost:8080/ServiceProviderConfig`

In each document we have to provide a notice, how the URLs would look like, when deploying in Tomcat.

tkrille:
Fixes #224


dacrome:
Yeah! Thanks a lot!

PR #238, opened 2016-06-11
Only deploy the WAR file without building again

tkrille:
Remove distribution management, because the repo id and URL is now
contained in the `circle.yml`. Remove sources plug-in, as we don't
distribute the sources on that way.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -26,7 +26,11 @@ deployment:
     branch: master
     commands:
       - wget https://raw.githubusercontent.com/osiam/circleci/master/settings.xml
-      - mvn deploy -s settings.xml -DskipTests
+      - >
+        mvn deploy:deploy-file
+          -DrepositoryId=snapshots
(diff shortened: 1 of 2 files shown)


circle.yml
```diff
@@ -36,5 +40,9 @@ deployment:
     owner: osiam
     commands:
       - wget https://raw.githubusercontent.com/osiam/circleci/master/settings.xml
-      - mvn deploy -s settings.xml -DskipTests
+      - >
+        mvn deploy:deploy-file
+          -DrepositoryId=releases
+          -Durl=https://api.bintray.com/maven/osiam/OSIAM/org.osiam:osiam/;publish=1
+          -s ../circleci/settings.xml
```
dacrome: Really this path? I think it is just `settings.xml`.

tkrille:
Updated

dacrome:
Great! Thanks a lot!

PR #243, opened 2016-06-19
Explain migration from PATCH to PUT

tkrille:


docs/migration.md
```diff
@@ -296,7 +296,45 @@ new OsiamConnector.Builder()
 
 #### API Changes
 
-There have been some minor API changes, namely:
+OSIAM 3.0 removes the obsolete PATCH API for updating resources.
+From now on, the only way to change a resource is to send a complete resource representation via PUT.
+This means, that you have to retrieve the complete resource first.
```
wallner: No comma, I think

docs/migration.md
```diff
@@ -296,7 +296,45 @@ new OsiamConnector.Builder()
...
+
+    becomes
+
+    ```java
+    new User.Builder(originalUser)
+            .setActive(false)
+            .build();
+    ```
+
+    Just pass the original resource, that you retrieved in step 1., to the builder's copy-of constructor.
```
wallner: The first comma is wrong.

tkrille: It actually is right, because this is a non-restrictive relative clause, but the pronoun is wrong. It should be "which". I've changed that.

docs/migration.md
tkrille: You're right. Thanks.

tkrille:
Thanks for your comments @wallner. I've updated the PR.

wallner:
Tank you!

PR #244, opened 2016-07-06
Remove the ability to update resources via PATCH

### LINKED ISSUE #21: Remove PATCH'ing of resources
At this time, we only implement the old PATCH API, the predecessor of the current JSON PATCH based API. It appears to me, that we don't really want to migrate to the new PATCH API right now, basically because of the high efforts that come with this undertaking. Due to this, I propose that we deprecate and eventually remove the old PATCH API. This way we also get rid of the not-so-beloved UpdateUser/Group. Components that have to be changed to achieve this:
- OSIAM
- connector4java
- addon-self-administration
- addon-administration (?)
- ITs and all other tests

So in fact, almost every component has to be changed!

tkrille:
Only remove the REST endpoints in the controllers. Everything else will
be removed in a future update. This approach has been chosen, to hold
the deadline for OSIAM 3.0. Removing the residues of updates via PATCH
is already in progress and will come with OSIAM 3.1.

Resolves #21


dacrome:
Please update `erviceProviderConfigController` that patch is not supported anymore.

tkrille:
Thanks for the hint. Updated.

dacrome:
Yeah!

PR #248, opened 2016-07-20
Bump dependencies and plugins

dacrome:


pom.xml
```diff
@@ -148,13 +148,13 @@
...
             <artifactId>postgresql</artifactId>
-            <version>9.4.1208</version>
+            <version>9.4.1209</version>
         </dependency>
 
         <dependency>
             <groupId>mysql</groupId>
             <artifactId>mysql-connector-java</artifactId>
-            <version>5.1.38</version>
+            <version>6.0.3</version>
```
tkrille: According to https://dev.mysql.com/downloads/connector/j/, version 6.0.3 is still a development release. The latest GA release is 5.1.39.

pom.xml
```diff
@@ -436,15 +436,15 @@
                 <artifactId>maven-deploy-plugin</artifactId>
                 <version>2.8.2</version>
                 <configuration>
-                    <file>${build.directory}/${project.artifactId}.war</file>
+                    <file>${project.build.directory}/${project.artifactId}.war</file>
```
tkrille, line 439: This doesn't really belong into this commit, does it?

pom.xml
```diff
@@ -458,19 +458,6 @@
...
-                <artifactId>maven-deploy-plugin</artifactId>
-                <version>2.8.2</version>
-                <dependencies>
-                    <dependency>
-                        <groupId>org.apache.maven.wagon</groupId>
-                        <artifactId>wagon-ssh-external</artifactId>
-                        <version>2.8</version>
-                    </dependency>
-                </dependencies>
-            </plugin>
```
tkrille, line 471: This doesn't really belong into this commit, does it?

pom.xml
dacrome: Ups. Changed to `5.1.39`.

pom.xml
dacrome, line 439: No. Added an own commit.

pom.xml
dacrome, line 471: No. Added an own commit.

PR #259, opened 2016-09-16
Remove wrong information from docs

tkrille:
Extensions can and should be created within the `osiam.yaml`.


dacrome:
Good catch! Thanks!

PR #260, opened 2016-09-25
Update Spring Boot to version 1.4.1

### LINKED ISSUE #258: Update to Spring Boot 1.4.1
I started [with the update in my fork](https://github.com/dacrome/osiam/tree/switch-to-spring-boot-1.4.0).

tkrille:
Remove `@Lob` and change `@Type` to `text` for text fields with variable
length. Hibernate 5 removed the type `StringClobType` and `text` is the
best replacement, as it does not map Strings to CLOBs, which does not work
with PotgreSQL.

Explicitly name some columns, as the naming strategy changed again in
Hibernate 5 and changing the column names in the database schema is rather
hard to do. This commit's scope is to only update Spring Boot.

Remove or replace deprecated things like
`spring.jpa.hibernate.naming_strategy` configuration property or
`org.springframework.boot.context.web.SpringBootServletInitializer`.

Also, do some code cleanup, use Spring Boot's auto configuration for
Flyway, and include Spring Boot devtools support. See individual commits
for details.

Resolves #258

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -220,6 +220,12 @@
         </dependency>
 
         <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-devtools</artifactId>
+            <optional>true</optional>
+        </dependency>
+
+        <dependency>
             <groupId>org.codehaus.groovy</groupId>
             <artifactId>groovy-all</artifactId>
             <version>2.4.7</version>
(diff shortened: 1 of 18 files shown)


tkrille:
@wallner: Because you do the Gradle migration, please notice that I've changed the POM with this PR.

dacrome:
Thank you very much!

Issue #263, opened 2016-10-05
The clientEntity id  is limited 32 characters 


wallner:
The client entity id is limited to 32 characters which is unnecessarily short and prevents the usage of, for instance, UUIDs as client names. I suggest raising this limit to at least 256 characters, or drop it all together.

dacrome:
Good catch! Let's do it!

tkrille:
+1

tkrille:
Added to 3.1

PR #264, opened 2016-10-06
Increase clientEntity id length to 256 characters

raj-subhankar:
Raised clientEntity id limit to at 256 characters. Issue #263.


dacrome:
Hey @raj-subhankar 

thanks for Pull Request!
You just forgot to add a new version of a [flyway](https://flywaydb.org/) migration which should looks like [this](https://github.com/osiam/osiam/blob/master/src/main/java/db/migration/V2__migrations_from_2_5.java). The migration has to change the length of the varchar with a SQL statement.

Cheers,
David

dacrome:
Hey @raj-subhankar,

what about my comment? Do you think you can add a new flyway migration to this PR?

Greetz,
David

raj-subhankar:
@dacrome hey sorry for the late reply. I am sorry, I don't know how to do that.

wallner:
Hello @raj-subhankar,
sorry for not answering earlier. I created a pull request based on yours with #278 that also adjusts all the needed database migrations. Thank you for your work!

PR #265, opened 2016-10-09
Introduce monitoring

### LINKED ISSUE #8: Monitoring of the OSIAM components
We should provide monitoring and metrics via the [Spring Boot Actuator](http://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#production-ready) for all OSIAM backend components. For sure this can only be done, if the components are migrated to Spring Boot.

dacrome:
- Add /monitor endpoint which groups a lot of endpoints [1] which
  provide monitoring information about the running OSIAM app
- Extends response of /monitor/info to provide java version and encoding
- Create short documentation
- Remove old /Metrics endpoint and dependencies

[1] http://docs.spring.io/spring-boot/docs/1.4.1.RELEASE/reference/html/production-ready-endpoints.html

Resolves #8


src/main/resources/application.yml
```diff
@@ -37,6 +37,13 @@ spring.thymeleaf.prefix: file:${osiam.home}/templates/web/
 spring.jpa.hibernate.ddl-auto: none
 server.port: ${osiam.port:8080}
 
+management.context-path: /monitor
```
dacrome: I'm not really happy with this path. Some more ideas: `/admin`, `/manage`, `/management` and `/monitoring`. Any other?

tkrille: It's more than monitoring, that's right. I would say `/manage` or `/management`. Take your pick! :)

CHANGELOG.md
```diff
@@ -1,5 +1,18 @@
 # OSIAM
 
+## 3.1 - 201x-xx-xx
```
tkrille: It's either "Unreleased" or no version at all. But "201x-xx-xx" looks kinda weird :)

src/main/resources/application.yml
```diff
@@ -37,6 +37,13 @@ spring.thymeleaf.prefix: file:${osiam.home}/templates/web/
...
+info.app:
+  encoding: @project.build.sourceEncoding@
+  java:
+    source: @java.version@
+    target: @java.version@
```
tkrille: I don't think we need to expose this kind of information. [This](http://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#production-ready-application-info-build) looks far more interesting. But it's better to add it, after the Gradle migration is done.

tkrille (changes requested):

tkrille:
Doesn't that break the ITs?

tkrille:
Indeed it does:

```
can_access_metrics_endpoint(org.osiam.client.AdminScopeIT)  Time elapsed: 0.893 sec  <<< FAILURE!
java.lang.AssertionError: 

Expected: is <200>
     but: was <404>
```

CHANGELOG.md
dacrome: Okay

src/main/resources/application.yml
dacrome: `management`

src/main/resources/application.yml
dacrome: Okay

dacrome:
Ups, sorry. I will update the ITs.

src/main/resources/application.yml
tkrille: :+1:

wallner: management is fine.

tkrille (approved):

Issue #256, opened 2016-09-03
Ability to plug the OSIAM as a decoupled component to SCIM user manager for accessing user core 


Vindulamj:
Hi @wallner 

I am trying to use OSIAM as a SCIM implementation which can be plugged into my own SCIM user manager which then talks to underlying user core using the user core API. For achieving this requirement what are the available possibilities within the OSIAM architecture which I can use ? . (eg - my SCIM user manager can implement the OSIAM API interface if such interface is available) 

Following diagram shows a high level overview of what my architecture need to be.
![12](https://cloud.githubusercontent.com/assets/12431869/18223997/71919870-71e5-11e6-9e2e-6868bfdb6124.jpeg)

Kindly do ask if anything is unclear or if I need to explain it more .

Regards
-Vindula

wallner:
Hello Vindula,
we don't really provide an interface like the one you are suggestinng yet, but I'm starting to think that it might be a good idea to do so. Your scenario looks like an interesting use case to me. We do however realize something lik you are thinking of with the LDAP mapping, albeit it is way cruder and clunkier than we like it to be and consider it something ready for refactoring. Please take a look at  `org.osiam.auth.login.ldap.OsiamLdapUserContextMapper` and tell me if that would be something like you had in mind.

tkrille:
Hi @Vindulamj,

SCIM 2.0 provides a [REST API](https://tools.ietf.org/html/rfc7644), which OSIAM implements. So, if your SCIM user manager has a generic client for this API, then you can just plug-in OSIAM or any other SCIM 2.0 compatible application. The only thing you have to consider is that you need to provide an access token for any request to OSIAM. You can get an access token with any generic OAuth 2 client as long as you configure the right URLs to the respective OAuth 2 endpoints. Your SCIM user manager must also know how the access token is structured. Consider using the [Connector4Java](https://github.com/osiam/connector4java/), if your application runs on the JVM.

BTW: I don't know, if I really understood your use case. But I think you want to integrate a SCIM 2.0 compatible service with your User Core service like you integrated a relational database (JDBC), LDAP and AD. To achieve this, you create the SCIM user manager, which is an adapter from User Core API to SCIM 2.0 API. If my understanding is wrong, please correct me and write some more details of your use case.

Vindulamj:
Hi @wallner @tkrille 

[...]

PR #270, opened 2016-10-25
Migrate build process to gradle

wallner:
This pull request swaps out the build process based on Maven and replaces it with a Gradle based one. This makes modifying the buildfile less painful and speeds up the build considerably.


tkrille:
@wallner Please rebase now.

tkrille:
Why no `gradlew.bat`? One should also be able to build OSIAM on Windows, shouldn't they?

tkrille:
If I run `gradle wrapper` or `./gradlew wrapper` I get a different `gradlew` script. It appears to be newer, as it contains an additional bug fix and supports an additional environment. The `gradle-wrapper.jar` also differs. Please run `./gradlew wrapper` again and check if this changes the aforementioned files too.

build.gradle
```diff
@@ -0,0 +1,187 @@
...
+    id 'maven-publish'
+    id 'com.github.kt3k.coveralls' version '2.6.3'
+    id 'com.jfrog.bintray' version '1.7.1'
+}
+
+apply plugin: 'spring-boot'
+
+group = 'org.osiam'
+
+description = """OSIAM"""
```
tkrille: Description in POM was:

```
OSIAM is a secure identity management solution providing REST based services for authentication and
authorization. We achieve this by implementing two important open standards:

* OAuth 2.0
* SCIM 2.0

OSIAM is published under the MIT licence, giving you the greatest freedom possible to utilize OSIAM in you
project or product.
```

For what is the description used anyway? Maybe we can leave it off altogether?

build.gradle
```diff
@@ -0,0 +1,187 @@
...
+        vcsUrl = 'https://github.com/osiam/osiam.git'
+        version {
+            name = project.version
+            released = new Date()
+        }
+    }
+
+}
+war {
+    baseName = 'OSIAM'
```
tkrille: This must be `osiam`.

build.gradle
```diff
@@ -0,0 +1,187 @@
...
+    jpaModelGen
+}
+
+dependencies {
+    compile 'org.springframework.boot:spring-boot-starter-web'
+    compile 'org.springframework.boot:spring-boot-starter-data-jpa'
+    compile 'org.springframework.boot:spring-boot-starter-thymeleaf'
+    compile 'org.springframework.boot:spring-boot-starter-security'
+    compile 'org.springframework.security.oauth:spring-security-oauth2'
+    compile 'org.springframework.boot:spring-boot-starter-tomcat'
```
tkrille: This must be removed. It is defined again below with the right scope.

build.gradle
```diff
@@ -0,0 +1,187 @@
...
+    testCompile 'org.codehaus.groovy:groovy-all'
+    testCompile 'org.springframework.boot:spring-boot-starter-test'
+    testCompile 'org.spockframework:spock-core'
+    testCompile 'cglib:cglib-nodep:3.2.4'
+    testCompile 'org.objenesis:objenesis:2.4'
+    testCompile 'com.jayway.restassured:json-path:2.9.0'
+    testCompile 'com.jayway.jsonpath:json-path:2.2.0'
+    testCompile 'nl.jqno.equalsverifier:equalsverifier:1.7.8'
+
+    providedRuntime("org.springframework.boot:spring-boot-starter-tomcat")
```
tkrille: Shouldn't this be `providedRuntime 'org.springframework.boot:spring-boot-starter-tomcat'`, like the other entries?

tkrille:
I cannot find a replacement for `com.mycila:license-maven-plugin`.

tkrille:
I cannot find a replacement for `org.pitest:pitest-maven`.

tkrille:
I'm missing a `classes` JAR, i.e. a normal JAR file that contains alle classes of OSIAM. It was created before with this configuration in the POM: https://github.com/osiam/osiam/blob/master/pom.xml#L488

tkrille:
The WAR file previously contained `CHANGELOG.md` and `README.md`. I'm missing these files now.

wallner:
It was not my intention to build a one-to-one replacement for the maven build. I think, for instance, that dropping the fuzzing for now is a good thing, since we never really used it anyways. I could be convinced to add a license plugin, is that something we want?

tkrille:
Alright, fair enough. We can drop the Pitest plugin, as we never used it anyways.

Regarding the license plugin: If we want each file to begin with a license header, then we should automate it (which implies enforcing it). I'm not a big fan of a license header in each file, but it seems common practice. So I would say, that we add a similar plugin.

Regarding the `classes` JAR: This was once requested by @jvmlet to make it possible to embed OSIAM in a custom Spring app. We also offered the possibility to do that in https://github.com/osiam/osiam/issues/256. I'm not comfortable with just removing this artifact. At least, it must be listed in the changelog that "feature" was removed.

Regarding the `CHANGELOG.md` and `README.md` in the WAR: The WAR file is our only distributable artifact, so I think it must contain at least the changelog. Having the readme, too, is also a good starting point for users. If you think that this is too much clutter, then at least mention the removal in the changelog. If users need it, they will eventually redemand it.

circle.yml
```diff
@@ -4,37 +4,32 @@ machine:
 
 dependencies:
   override:
-    - mvn -U dependency:resolve dependency:resolve-plugins
```
tkrille, line 7: Is there any replacement for this line? It would be nice to cache downloaded dependencies and build plugins.

circle.yml
```diff
@@ -4,37 +4,32 @@ machine:
...
   post:
     - mkdir -p $CIRCLE_TEST_REPORTS/junit/
-    - find . -type f -regex ".*/target/.*-reports/.*xml" -exec cp {} $CIRCLE_TEST_REPORTS/junit/ \;
-    - cp -r target/coverage-reports/jacoco/ $CIRCLE_ARTIFACTS
-    - test -z $CODACY_PROJECT_TOKEN || codacy-coverage-reporter -l Java -r target/coverage-reports/jacoco/jacoco.xml --projectToken $CODACY_PROJECT_TOKEN
-    - test -z $COVERALLS_REPO_TOKEN || mvn coveralls:report -DrepoToken=$COVERALLS_REPO_TOKEN
+    - find . -type f -regex ".*/build/test-results/*.*xml" -exec cp {} $CIRCLE_TEST_REPORTS/junit/ \;
+    - cp -r build/reports/jacoco/ $CIRCLE_ARTIFACTS
+    - test -z $CODACY_PROJECT_TOKEN || codacy-coverage-reporter -l Java -r build/reports/jacoco/test/jacocoTestReport.xml --projectToken $CODACY_PROJECT_TOKEN
+    - test -z $COVERALLS_REPO_TOKEN || ./gradlew jacocoTestReport -DrepoToken=$COVERALLS_REPO_TOKEN
```
tkrille: I think the task must be `coveralls`, not `jacocoTestReport`. And is `-DrepoToken=$COVERALLS_REPO_TOKEN` really needed? It runs perfectly fine on my machine with just the environment variable defined.

circle.yml
```diff
@@ -4,37 +4,32 @@ machine:
...
   override:
-    - mvn verify
+    - test -z ${CIRCLE_TAG} && ./gradlew check || ./gradlew -Pversion=${CIRCLE_TAG} check
   post:
     - mkdir -p $CIRCLE_TEST_REPORTS/junit/
-    - find . -type f -regex ".*/target/.*-reports/.*xml" -exec cp {} $CIRCLE_TEST_REPORTS/junit/ \;
-    - cp -r target/coverage-reports/jacoco/ $CIRCLE_ARTIFACTS
-    - test -z $CODACY_PROJECT_TOKEN || codacy-coverage-reporter -l Java -r target/coverage-reports/jacoco/jacoco.xml --projectToken $CODACY_PROJECT_TOKEN
-    - test -z $COVERALLS_REPO_TOKEN || mvn coveralls:report -DrepoToken=$COVERALLS_REPO_TOKEN
+    - find . -type f -regex ".*/build/test-results/*.*xml" -exec cp {} $CIRCLE_TEST_REPORTS/junit/ \;
```
tkrille: The Circle CI docs state that it must be

```
- find . -type f -regex ".*/build/test-results/.*xml" -exec cp {} $CIRCLE_TEST_REPORTS/junit/ \;
```

I.e. there's a `*` too much before `.*xml`.

circle.yml
```diff
@@ -4,37 +4,32 @@ machine:
...
       - >
         curl -H "Content-Type: application/json" --data '{"source_type": "Branch", "source_name": "master"}' -X POST https://registry.hub.docker.com/u/osiamorg/osiam/trigger/${DOCKER_HUB_TRIGGER_TOKEN}/
   release:
     tag: /.*/
     owner: osiam
     commands:
-      - wget https://raw.githubusercontent.com/osiam/circleci/master/settings.xml
-      - mvn deploy:deploy-file -DrepositoryId=releases -Durl="https://api.bintray.com/maven/osiam/OSIAM/org.osiam:osiam/;publish=1" -s settings.xml
-      - curl -T target/osiam.war -u${BINTRAY_USER}:${BINTRAY_KEY} -H "X-Bintray-Package:osiam" -H "X-Bintray-Version:${CIRCLE_TAG}" -H "X-Bintray-Publish:1" -H "X-Bintray-Override:1" https://api.bintray.com/content/osiam/downloads/osiam/${CIRCLE_TAG}/osiam-${CIRCLE_TAG}.war
+      - ./gradlew -Pversion=${CIRCLE_TAG} bintrayUpload
```
tkrille, line 36: Using this task doesn't automatically publish the uploaded artifacts. Please make sure that artifacts are published.

src/main/antlr/org/osiam/storage/parser/LogicalOperatorRules.g4
```diff
@@ -24,6 +24,10 @@
 /** Define a grammer called LogicalOperatorRules which will be the basis for ARNTL to generate the java classes for lexer and parser */
 grammar LogicalOperatorRules;
 
+@header {
+package org.osiam.storage.parser;
+}
```
tkrille, line 29: Why is this suddenly needed?

build.gradle
```diff
@@ -0,0 +1,185 @@
...
+}
+
+dependencies {
+    compile 'org.springframework.boot:spring-boot-starter-web'
+    compile 'org.springframework.boot:spring-boot-starter-data-jpa'
+    compile 'org.springframework.boot:spring-boot-starter-thymeleaf'
+    compile 'org.springframework.boot:spring-boot-starter-security'
+    compile 'org.springframework.boot:spring-boot-starter-actuator'
+    compile 'org.springframework.security.oauth:spring-security-oauth2'
+    compile 'org.springframework.boot:spring-boot-starter-tomcat'
```
tkrille: This must be removed. It is defined again below with the right scope.

wallner:
Ok, I added the `gradlew.bat` and updated the wrapper. I'm ok with adding a license plugin, however I don't really see a value in adding the `CHANGELOG.md` and `README.md` to the war-file. Do you really believe people are extracting that? Somehow I doubt it.

build.gradle
```diff
@@ -0,0 +1,185 @@
...
+        licenses = ['MIT']
+        vcsUrl = 'https://github.com/osiam/osiam.git'
+        version {
+            name = project.version
+            released = new Date()
+        }
+    }
+
+}
+war {
```
tkrille, line 90: Is there an easy way to just name the resulting WAR file `osiam.war`. This was the previous behavior and I found it kinda convenient.

tkrille (changes requested):

tkrille:
> however I don't really see a value in adding the `CHANGELOG.md` and `README.md` to the war-file. Do you really believe people are extracting that? Somehow I doubt it.

Yeah, I'm not so sure anymore either. When in doubt, we don't do it. Let's leave it off.

dacrome:
hm, but the war is our released artifact. If it's no big deal, please add it.

build.gradle
wallner: Let's drop the description. I agree.

circle.yml
wallner: Well, then it was wrong in the first place.

circle.yml
wallner: I don't think so.

circle.yml
wallner: Probably not, but I tried to change as little here as possible, so I left the Property untouched. Removed now.

src/main/antlr/org/osiam/storage/parser/LogicalOperatorRules.g4
wallner: The Gradle antlr plugin does a little less magic than the maven one. Which I consider a good thing.

circle.yml
tkrille: No, it wasn't. The previous version was

```
find . -type f -regex ".*/target/.*-reports/.*xml" -exec cp {} $CIRCLE_TEST_REPORTS/junit/ \;
```

which contains a correct regex. Nevertheless, please remove the superfluous `*`.

src/main/antlr/org/osiam/storage/parser/LogicalOperatorRules.g4
tkrille: Alright, thanks for clarification.

circle.yml
tkrille: Please check the [CircleCI docs](https://circleci.com/docs/language-java/#dependencies--tests): It is `./gradlew dependencies`.

wallner:
I don't think it make sense to include `README.md` and `CHANGELOG.md`. I can't see a reason for it. Do you really think people are unpacking the war file? I don't. I know at least I wouldn't. If you insist on providing an artefact that contains both files, I suggest that we build one archive that contains the `war` and the two files separately.

dacrome:
Yes, you are right. Leave it off.

wallner:
I found a license-plugin replacement and added it.

build.gradle
```diff
@@ -0,0 +1,195 @@
...
+}
+
+compileJava {
+    dependsOn generateJPAMetaModel
+}
+
+license {
+    mapping {
+        g4 = 'JAVADOC_STYLE'
+        excludes (["**/*_.java", "**/LogicalOperatorRules*.java", "**/*.json"])
```
tkrille: This line must be outside the closure for `mapping`. Though it works, it does so merely by accident.

circle.yml
```diff
@@ -4,37 +4,33 @@ machine:
...
 
 deployment:
   staging:
     branch: master
     commands:
-      - wget https://raw.githubusercontent.com/osiam/circleci/master/settings.xml
-      - mvn deploy:deploy-file -DrepositoryId=snapshots -Durl=https://oss.jfrog.org/artifactory/oss-snapshot-local -s settings.xml
-      - curl -T target/osiam.war -u${BINTRAY_USER}:${BINTRAY_KEY} -H "X-Bintray-Package:osiam" -H "X-Bintray-Version:latest" -H "X-Bintray-Publish:1" -H "X-Bintray-Override:1" https://api.bintray.com/content/osiam/downloads/osiam/latest/osiam-latest.war
+      - ./gradlew publish
+      - curl -T build/libs/OSIAM.war -u${BINTRAY_USER}:${BINTRAY_KEY} -H "X-GPG-PASSPHRASE:${GPG_PASSPHRASE}" -H "X-Bintray-Package:osiam" -H "X-Bintray-Version:latest" -H "X-Bintray-Publish:1" -H "X-Bintray-Override:1" https://api.bintray.com/content/osiam/downloads/osiam/latest/osiam.war
```
tkrille: This must be `curl -T build/libs/osiam.war -u${BINTRAY_USER} …` now. The WAR file is lower-cased again.

tkrille: The URL to POST to must be `https://api.bintray.com/content/osiam/downloads/osiam/latest/osiam-latest.war` like before.

tkrille (changes requested):

tkrille:
Alright, looks pretty good to me now. There are 3 little things remaining. Please fix them and I'll merge this immediately.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,6 @@
+#Thu Oct 27 23:26:06 CEST 2016
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+zipStoreBase=GRADLE_USER_HOME
+zipStorePath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-3.1-all.zip
```
dacrome, line 6: Can you use Gradle 3.2?

circle.yml
```diff
@@ -4,37 +4,33 @@ machine:
...
+    - test -z $COVERALLS_REPO_TOKEN || ./gradlew coveralls
 
 deployment:
   staging:
     branch: master
     commands:
-      - wget https://raw.githubusercontent.com/osiam/circleci/master/settings.xml
-      - mvn deploy:deploy-file -DrepositoryId=snapshots -Durl=https://oss.jfrog.org/artifactory/oss-snapshot-local -s settings.xml
-      - curl -T target/osiam.war -u${BINTRAY_USER}:${BINTRAY_KEY} -H "X-Bintray-Package:osiam" -H "X-Bintray-Version:latest" -H "X-Bintray-Publish:1" -H "X-Bintray-Override:1" https://api.bintray.com/content/osiam/downloads/osiam/latest/osiam-latest.war
+      - ./gradlew publish
```
tkrille: This will upload a WAR that has not been enhanced and repackaged by Spring Boot. Either make this `./gradlew assemble publish` or let the `publish` task depend on `assemble`.

circle.yml
```diff
@@ -4,37 +4,33 @@ machine:
...
       - >
         curl -H "Content-Type: application/json" --data '{"source_type": "Branch", "source_name": "master"}' -X POST https://registry.hub.docker.com/u/osiamorg/osiam/trigger/${DOCKER_HUB_TRIGGER_TOKEN}/
   release:
     tag: /.*/
     owner: osiam
     commands:
-      - wget https://raw.githubusercontent.com/osiam/circleci/master/settings.xml
-      - mvn deploy:deploy-file -DrepositoryId=releases -Durl="https://api.bintray.com/maven/osiam/OSIAM/org.osiam:osiam/;publish=1" -s settings.xml
-      - curl -T target/osiam.war -u${BINTRAY_USER}:${BINTRAY_KEY} -H "X-Bintray-Package:osiam" -H "X-Bintray-Version:${CIRCLE_TAG}" -H "X-Bintray-Publish:1" -H "X-Bintray-Override:1" https://api.bintray.com/content/osiam/downloads/osiam/${CIRCLE_TAG}/osiam-${CIRCLE_TAG}.war
+      - ./gradlew -Pversion=${CIRCLE_TAG} bintrayUpload
```
tkrille, line 36: See https://github.com/osiam/osiam/pull/270/files#r88811799

build.gradle
wallner: That's fixed now.

circle.yml
```diff
@@ -4,37 +4,35 @@ machine:
...
 
 deployment:
   staging:
     branch: master
     commands:
-      - wget https://raw.githubusercontent.com/osiam/circleci/master/settings.xml
-      - mvn deploy:deploy-file -DrepositoryId=snapshots -Durl=https://oss.jfrog.org/artifactory/oss-snapshot-local -s settings.xml
-      - curl -T target/osiam.war -u${BINTRAY_USER}:${BINTRAY_KEY} -H "X-Bintray-Package:osiam" -H "X-Bintray-Version:latest" -H "X-Bintray-Publish:1" -H "X-Bintray-Override:1" https://api.bintray.com/content/osiam/downloads/osiam/latest/osiam-latest.war
+      - ./gradlew assemble
+      - ./gradlew publish
```
tkrille: Unfortunately, `publish` depends on `war`, because of the `from components.web` statement in the publication's definition. Because the output of the original `war` task has changed (Spring Boot repackaged the WAR file), the `war` task will run again, but this time without the repackaging. This produces a broken WAR file that gets published. I'd suggest to merge the 2 lines: `./gradlew assemble publish`. This way, the repackaged WAR file is published. Please note that this is not true for the `bintrayUpload` task. So the next line can stay as it is.

circle.yml
```diff
@@ -4,37 +4,35 @@ machine:
...
         curl -H "Content-Type: application/json" --data '{"source_type": "Branch", "source_name": "master"}' -X POST https://registry.hub.docker.com/u/osiamorg/osiam/trigger/${DOCKER_HUB_TRIGGER_TOKEN}/
   release:
     tag: /.*/
     owner: osiam
     commands:
-      - wget https://raw.githubusercontent.com/osiam/circleci/master/settings.xml
-      - mvn deploy:deploy-file -DrepositoryId=releases -Durl="https://api.bintray.com/maven/osiam/OSIAM/org.osiam:osiam/;publish=1" -s settings.xml
-      - curl -T target/osiam.war -u${BINTRAY_USER}:${BINTRAY_KEY} -H "X-Bintray-Package:osiam" -H "X-Bintray-Version:${CIRCLE_TAG}" -H "X-Bintray-Publish:1" -H "X-Bintray-Override:1" https://api.bintray.com/content/osiam/downloads/osiam/${CIRCLE_TAG}/osiam-${CIRCLE_TAG}.war
+      - ./gradlew -Pversion=${CIRCLE_TAG} assemble
+      - ./gradlew -Pversion=${CIRCLE_TAG} publish
```
tkrille: See https://github.com/osiam/osiam/pull/270/files#r89203451.

tkrille (changes requested):

tkrille:
Besides my last two comments, everything works as expected.

circle.yml
wallner: Thank you for looking into this.

tkrille:
Thank you very much!

PR #271, opened 2016-10-26
Fix typo in the initially created OAuth 2 client's client ID

andreyakostov:
The client ID actually has a hyphen between the words example and client.


wallner:
Good Catch! Thank you for fixing it!

andreyakostov:
You're welcome. :)

PR #278, opened 2016-12-03
Increase the length of client id to 255 characters

### LINKED ISSUE #263: The clientEntity id  is limited 32 characters

wallner:
The length of the scim client id was limited to 32 characters. That was
a little short to be useful. This commit inreases the length to more
useful 255 characters.

Resolves #263 closes #264

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 ### Changes
 
 - Remove `/Metrics` endpoint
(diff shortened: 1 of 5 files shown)


tkrille:
Thanks

PR #286, opened 2017-02-10
Update Gradle and Spring Boot

dacrome:


dacrome:
I need to run the ITs before this PR can be merged

tkrille:
Please apply this patch to the `application.yml`:

```
diff --git a/src/main/resources/application.yml b/src/main/resources/application.yml
index 39d36bf..abf6d6c 100644
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -36,6 +36,7 @@ spring.main.banner-mode: 'off'
 spring.thymeleaf.prefix: file:${osiam.home}/templates/web/
 spring.jpa.hibernate.ddl-auto: none
+security.oauth2.resource.filterOrder: 3
 
 management.context-path: /management
```

tkrille:
But we still have to think about the order of the security filter chains.

dacrome:
That is the fix, thanks @tkrille

PR #294, opened 2017-04-11
Update Gradle to version 3.5

dacrome:


dacrome:
http://www.jpm4j.org/install/script is currently not available that's why the build fails.

tkrille:
> http://www.jpm4j.org/install/script is currently not available that's why the build fails.

I don't think it will ever be available again. We should replace it.

PR #303, opened 2017-05-28
Some improvements

tkrille:


tkrille:
I've fixed the build errors

PR #305, opened 2017-05-30
Store session data in relational database

### LINKED ISSUE #293: Persist sessions
* With Spring Session
* Initial support for JDBC
* JDBC is default
* Support for Hazelcast and Redis will be added later on

wallner:
This pull request uses spring-session-jdbc to store the session data in the relational database. This is a first shot and we might replace the relational database with something better suited like Hazelcast or Redis.

Resolves #293

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -106,6 +106,7 @@ dependencies {
     compile 'org.springframework.boot:spring-boot-starter-actuator'
     compile 'org.springframework.security.oauth:spring-security-oauth2'
     compile 'org.springframework.security:spring-security-ldap'
+    compile 'org.springframework.session:spring-session:1.3.1.RELEASE'
     compile 'org.springframework.boot:spring-boot-devtools'
     compile 'org.flywaydb:flyway-core'
     compile 'org.postgresql:postgresql'
--- a/src/main/java/org/osiam/Osiam.java
+++ b/src/main/java/org/osiam/Osiam.java
@@ -37,6 +37,7 @@ import org.springframework.context.annotation.FilterType;
 import org.springframework.security.authentication.encoding.ShaPasswordEncoder;
 import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
 import org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder;
+import org.springframework.session.jdbc.config.annotation.web.http.EnableJdbcHttpSession;
 
 import java.util.Arrays;
 import java.util.Collections;
@@ -44,6 +45,7 @@ import java.util.Map;
 
 @SpringBootApplication
 @EnableWebSecurity
+@EnableJdbcHttpSession
 @EnableAspectJAutoProxy(proxyTargetClass = true)
 @ComponentScan(excludeFilters = {
         @ComponentScan.Filter(type = FilterType.REGEX, pattern = "org\\.osiam\\.cli.*")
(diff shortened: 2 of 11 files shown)


build.gradle
```diff
@@ -106,6 +106,7 @@ dependencies {
     compile 'org.springframework.boot:spring-boot-starter-actuator'
     compile 'org.springframework.security.oauth:spring-security-oauth2'
     compile 'org.springframework.security:spring-security-ldap'
+    compile 'org.springframework.session:spring-session:1.3.1.RELEASE'
```
tkrille, line 109: `spring-session` is set to version 1.3.0.RELEASE by the Spring Boot 1.5.3 BOM. Any specific reason why we need the latest version beyond "because we want the latest version"?

src/main/java/org/osiam/Osiam.java
```diff
@@ -37,13 +37,15 @@
...
 import org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder;
+import org.springframework.session.jdbc.config.annotation.web.http.EnableJdbcHttpSession;
 
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Map;
 
 @SpringBootApplication
 @EnableWebSecurity
+@EnableJdbcHttpSession
```
tkrille, line 48: This is not needed. Spring Boot has an [Auto Configuration](https://github.com/spring-projects/spring-boot/blob/v1.5.3.RELEASE/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/SessionAutoConfiguration.java) for Spring Session. The Auto Configuration will also make it possible to [define the session timeout](https://github.com/spring-projects/spring-boot/blob/v1.5.3.RELEASE/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/JdbcSessionConfiguration.java#L61) via the [`spring.session` configuration property namespace](https://github.com/spring-projects/spring-boot/blob/v1.5.3.RELEASE/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/SessionProperties.java). But you have to set the configuration property `spring.session.store-type` to `jdbc` so that the Auto Configuration is picked up. See the [reference manual](http://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#boot-features-session) for details.

src/main/resources/application.yml
```diff
@@ -38,6 +38,12 @@ spring.jpa.hibernate.ddl-auto: none
 server.port: ${osiam.port:8080}
 security.oauth2.resource.filterOrder: 3
 
+spring:
+  datasource:
```
tkrille, line 42: Configuring the datasource is not needed. As we already define our [own `Datasource` Bean](https://github.com/osiam/osiam/blob/master/src/main/java/org/osiam/configuration/DatabaseConfiguration.java#L60), these values will never be picked up by the respective Auto Configuration.

src/main/java/org/osiam/auth/login/internal/InternalAuthenticationProvider.java
```diff
@@ -117,13 +113,11 @@ public Authentication authenticate(Authentication authentication) {
...
         }
 
-        User authUser = new User.Builder(username).setId(user.getId()).build();
-
         List<GrantedAuthority> grantedAuthorities = user.getRoles().stream()
                 .map(role -> new SimpleGrantedAuthority(role.getValue()))
                 .collect(Collectors.toList());
 
-        return new InternalAuthentication(authUser, password, grantedAuthorities);
+        return new InternalAuthentication(username, password, grantedAuthorities);
```
tkrille, line 120: Unfortunately, this does not work. The principal must be the `User`, because it's the base for generating the access tokens in the resource owner credentials grant and authorization code grant. I get a bunch of errors when running the ITs related to missing data in the access tokens.

tkrille (changes requested):

tkrille:
Just for my curiosity: Are expired sessions removed from the database and how is this done?

tkrille (changes requested):
Please add an entry to the changelog.

PR #307, opened 2017-08-01
Update Spring Boot and dependencies and plugins and Gradle

dacrome:


dacrome:
I was not able to run the ITs because I have not a compatible Docker version installed for the docker-maven-plugin: `com.github.dockerjava.api.exception.DockerClientException: Could not build image: null` - but CircleCi will tell use ;)

